PR #8, opened 2026-01-05
output(console): reduced output for constraints

matthiasneuner:


alexdummer (commented):
The only thing to discuss is if we should mark such changes with "feat" which in my opinion indicates a real new feature. Maybe we use "output(console)" for this purpose?

PR #10, opened 2026-01-19
format: add cython-lint and clang-format

alexdummer:


edelweissfe/linsolve/pardiso/pardiso.pyx
matthiasneuner, line 1: Several header get malformed?

alexdummer: this was my script to apply the requirements, now fixed!

PR #40, opened 2026-06-19
refactor: introduce structured 2-D views and wrappers for stiffness m…

matthiasneuner:
…atrices

Stiffness matrices for finite elements and constraints were previously passed around as flat 1-D arrays, forcing implementations to manually reshape them by hand or perform error-prone offset/stride math inside constraints.

This commit refactors how stiffness matrices are shaped and accessed:
- Introduces `shapeVIJContribution` in `VIJEntityBase` allowing each entity to define how its stiffness slice is structured (defaulting to a 2-D Fortran-order view of shape `(nDof, nDof)`).
- Updates `VIJSystemMatrix.__getitem__` to delegate shaping to the entity automatically.
- Implements `RigidBodyStiffnessView` in `rigidbody.py` to wrap the sparse flat matrix and expose properly sized 2-D views (`K_UU`, `K_UL`, and `K_LU`).
- Simplifies `Constraint.applyConstraint` in `rigidbody.py` to write directly into these 2-D views, eliminating manual flattening and indexing math.
- Simplifies the `assembleConstraints` loop in the solvers (removing the conditional reshaping logic).
- Simplifies `computeYourself` in elements to accept the 2-D views directly, and overrides shaping to return 1-D views for Cython-wrapped elements to preserve memoryview compatibility.


edelweissfe/constraints/rigidbody.py
```diff
@@ -53,6 +53,37 @@
 documentation = [module]
 
 
+class RigidBodyStiffnessView:
+    """Provides structured 2-D sub-views for the sparse rigid body stiffness matrix slice."""
```
alexdummer: maybe add a more precise documentation for this object?

edelweissfe/numerics/vijsystemmatrix.py
```diff
@@ -75,7 +75,10 @@ def __getitem__(self, key):
             # Entity Lookup
             idxInVIJ = self.entitiesInVIJ[key]
             size = key.getVIJContributionSize()
-            return super().__getitem__(slice(idxInVIJ, idxInVIJ + size))
+            flat_view = super().__getitem__(slice(idxInVIJ, idxInVIJ + size))
+            if hasattr(key, "shapeVIJContribution"):
```
alexdummer: is this necessary? every object should be of vijentitybase and automatically has the default shapeVIJContribution?

PR #41, opened 2026-06-19
chore: update element internal force sign and remove pNewDT and stepTime

matthiasneuner:
- Modified python elements (DisplacementElement, DisplacementTLElement, MarmotMaterialWrappingElement) to accumulate internal forces with a positive sign.
- Updated solvers to compute the residual as R = P_ext - P_int.
- Removed outdated pNewDT / pNewdT reference parameters from Cython element.pxd and element.pyx wrappers.
- Enabled standard exception translation (+except) and modified Python solvers to catch RuntimeError to trigger step cutbacks.

### DIFF
--- a/edelweissfe/solvers/nonlinearimplicitstatic.py
+++ b/edelweissfe/solvers/nonlinearimplicitstatic.py
@@ -229,7 +229,7 @@ class NIST(NonlinearSolverBase):
                         maxGrowingIter,
                     )
 
-                except (CutbackRequest, RuntimeError) as e:
+                except CutbackRequest as e:
                     self.journal.message(str(e), self.identification, 1)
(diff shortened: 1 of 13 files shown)


edelweissfe/solvers/nonlinearimplicitstatic.py
```diff
@@ -229,9 +229,10 @@ def solveStep(
                         maxGrowingIter,
                     )
 
-                except CutbackRequest as e:
+                except (CutbackRequest, RuntimeError) as e:
```
alexdummer: I think not every RuntimeError should result in a cutback? Can we use a more specific exception here?

alexdummer (approved):

PR #43, opened 2026-06-20
fix(docs): fix Sphinx documentation build failure by lazy-loading inputfileparser

matthiasneuner:
…parser

This resolves import-order dependencies that caused step modules (like adaptivestep.py) to be loaded before inputfileparser.py is initialized, preventing the step modules from registering themselves under the *step keyword.


advktEntnschdl (approved):
I don't have a better solution, so I'm fine with it, if it solves the problem.

alexdummer (approved):

PR #57, opened 2026-07-18
feat: node-to-surface contact with facets

matthiasneuner:


alexdummer (commented):
Automated code review comparing this branch against `next_v26.11`. 7 inline findings below on lines touched by this PR. 3 additional findings landed on code this PR doesn't modify (so GitHub won't let me attach them inline) -- posting those separately as a regular comment.

edelweissfe/generators/surfaceelementgenerator.py
```diff
@@ -0,0 +1,310 @@
...
+# Quad faces of 3D solids are split into two Tria3 facets via a diagonal, preserving the source
+# face's winding (and thus outward orientation).
+#
+# These follow this codebase's own local node numbering and face-number convention as generated by
+# edelweissfe.generators.boxgen/planerectquad (NOT the generic Abaqus S1..S6 convention, and NOT the
+# "obvious" bottom/top=Z axis reading either -- boxgen's own axis convention is bottom/top <-> Y,
+# left/right <-> X, front/back <-> Z; faces are numbered 1=Ymin, 2=Ymax, 3=Xmin, 4=Zmax, 5=Xmax,
+# 6=Zmin). Verified against boxgen's actual node layout (local node k <-> (ix, iy, iz) offsets) by
+# checking that each listed triangle's cross-product normal points outward from the source element.
+_FACE_TABLES = {
```
alexdummer, line 105: **Correctness:** `_FACE_TABLES`/`_MIDSIDE_FACE_TABLES` hardcode boxgen/planerectquad's own face-numbering convention, but `*surface` definitions parsed from a standard Abaqus-style input file (`abqmodelconstructor.py:240`, `faceNumber = int(S-label)`) use the real Abaqus S1..S6 convention -- which the comment above this table itself says differs.

**Failure scenario:** Importing an Abaqus-style hexa20 mesh and defining `*surface S3` on a hex face looks up `faceNumber=3` in `_FACE_TABLES['hexa20']`, which under boxgen's convention means a specific axis-aligned face -- not whatever Abaqus's real S3 numbering denotes for that mesh. The generator silently triangulates the wrong physical face into a contact facet, with no validation catching the mismatch.

edelweissfe/constraints/nodetodeformablesurfacepenalty.py
```diff
@@ -0,0 +1,827 @@
...
+                        bestDistance, bestFacet, bestWeights = distance, i, weights
+
+                if bestFacet is not None and (self.searchDistance is None or bestDistance <= self.searchDistance):
+                    newAssignment[s] = bestFacet
+                    self._frozenWeights[s] = bestWeights
+                    normal, _ = facetNormalAndMeasure(facetCoords[bestFacet])
+                    self._frozenNormals[s] = normal
+                    # Rotate the frictional history into the new frozen tangent plane; the frame
+                    # changes only slightly per increment in a small-deformation setting.
+                    projectorOntoTangentPlane = np.eye(self.nDim) - np.outer(normal, normal)
```
alexdummer: **Correctness:** `updateConnectivity()` unconditionally rotates the converged tangential-force history and Lagrange multiplier into the newly assigned facet's tangent plane, even when the new facet is unrelated to the previous one -- there's no reset on a facet change, only a comment assuming "the frame changes only slightly per increment."

**Failure scenario:** A slave node sliding across a sharp corner/crease of the master surface gets reassigned from facet A to a differently-oriented facet B within one increment. Lines 517-518 project the converged tangential force into facet B's tangent plane using facet B's normal, reusing stale friction/history from facet A as the stick predictor on facet B -- producing spurious tangential force/AL-multiplier carryover that nothing enforces is actually small.

edelweissfe/rigidbodies/discreterigidbody.py
```diff
@@ -97,18 +98,19 @@ def getCurrentKinematics(self):
...
-                "'displacement' NodeField -- the model may not be fully initialized yet."
-            )
-        u_rp = disp_field["U"][disp_field.indexOfNode(self.rpNode)].copy()
+        if disp_field is not None and "U" in disp_field and self.rpNode in disp_field.nodes:
+            idx = disp_field._indicesOfNodesInArray[self.rpNode]
+            u_rp = disp_field["U"][idx].copy()
+        else:
+            u_rp = np.zeros(self.model.domainSize)
 
         R = np.eye(3)
```
alexdummer, line 107: **Correctness:** `getCurrentKinematics()` only reads/applies the rigid-body rotation field when `model.domainSize == 3` (this `if` guard is new in this PR); for a 2D model, `R` stays `np.eye(3)` forever. Nothing in `DiscreteRigidBody`/`discreterigidbodygenerator.py` rejects 2D models -- only the separate `nodeToDiscreteRigidBodyPenalty` constraint explicitly raises `ValueError` for `domainSize != 3` (`nodetodiscreterigidbodypenalty.py:233-234`).

**Failure scenario:** A 2D discrete rigid body driven by a rotational BC or torque on its RP node silently never rotates -- `querySurface`, `updateKinematics`, and `getVisualizationField('displacement')` all use `R=I` forever, with no error.

edelweissfe/constraints/nodetodeformablesurfacepenalty.py
```diff
@@ -0,0 +1,827 @@
...
+
+            self._normalForceCurrent[s] = f_n
+            self.totalNormalForce += f_n
+            activeIdx += 1
+
+    def getNormalPressures(self) -> np.ndarray:
+        """The current per-slave normal contact pressures (positive in compression), ordered like
+        the surface element generator's ``<prefix>_nodes`` node set of the slave surface."""
+
+        return -self._normalForceCurrent / self.tributaryAreas
```
alexdummer: **Correctness:** `getNormalPressures()`/`getTangentialTractions()` divide by `self.tributaryAreas` with no guard against a zero entry. `tributaryAreas` is built by summing `nodalAreaShares` over each slave node's incident facets, initialized from `0.0` (lines 379-385).

**Failure scenario:** A degenerate (zero-area) slave facet from duplicated/collinear mesh nodes leaves `tributaryAreas[s] == 0.0` for that node; `getNormalPressures()` then computes `-0/0 = nan`, silently poisoning any downstream `fromExpression` field-output aggregate (sum/max) reading it -- no exception is raised anywhere in the constraint.

edelweissfe/rigidbodies/discreterigidbody.py
```diff
@@ -264,18 +265,11 @@ def getVisualizationField(self, fieldName: str) -> np.ndarray:
...
-            The name of the field. Currently only ``"displacement"`` is supported.
-
-        Raises
-        ------
-        ValueError
-            If ``fieldName`` is not ``"displacement"``.
+            The name of the field. Currently only ``"displacement"`` is supported;
+            any other field returns zeros.
         """
         if fieldName != "displacement":
```
alexdummer, line 274: **Robustness:** `getVisualizationField` only implements `fieldName == 'displacement'`; every other field name silently returns an all-zero array instead of raising an error.

**Failure scenario:** A `RigidBodyFieldOutput` requesting `'velocity'` or `'rotation'` on a discrete rigid body returns `np.zeros((len(surfaceNodes), domainSize))` with no warning, so the output/visualization silently shows a rigid body that never moves/rotates even though the underlying kinematics are correct -- easy to miss since there's no error to flag the unsupported field.

edelweissfe/constraints/nodetodeformablesurfacepenalty.py
```diff
@@ -0,0 +1,827 @@
...
+
+        newAssignment = [None] * self.nSlaves
+
+        if self.sliding == "small":
+            # Clamped closest-point search: the assigned facet is the one whose closed domain
+            # (interior/edges/vertices) is truly closest -- no dead zone at facet seams, and the
+            # clamped weights are non-negative by construction. Facet, weights, and normal are
+            # frozen for the whole increment, making the gap linear in the displacement DOFs.
+            closestPointFunction = _tria3ClosestPoint if self.nDim == 3 else _line2ClosestPoint
+            for s in range(self.nSlaves):
```
alexdummer, line 530: **Efficiency:** The small-sliding candidate search is a brute-force O(nSlaves x nMasterFacets) nested Python loop with no spatial acceleration structure, run once per increment in the Newton-loop-adjacent hot path.

**Failure scenario:** For a slave surface with 1,000 nodes and a master surface with 10,000 facets, this is 10 million pure-Python closest-point evaluations per increment. Unlike the rigid-body contact path (`DiscreteSurfaceQuery`, which reuses a VTK spatial locator), the deformable-surface path has no bounding-volume/KD-tree acceleration, so wall time scales quadratically as either surface is refined and will dominate simulation cost well before Newton-iteration cost does.

edelweissfe/solvers/nonlinearimplicitstatic.py
```diff
@@ -173,22 +140,77 @@ def solveStep(
...
         prevTimeStep = None
 
         self.applyStepActionsAtStepStart(model, step.actions)
 
         try:
             for timeStep in step.getTimeStep():
+                # NOTE: materialize the list before any() -- a generator would short-circuit at
+                # the first constraint reporting a change, silently skipping updateConnectivity()
+                # for every remaining constraint (their connectivity would stay stale/empty).
+                connectivityHasChanged = any(
```
alexdummer, line 163: **Efficiency:** Any single constraint reporting a connectivity change triggers a full `DofManager`/VIJ-sparsity-pattern/CSR rebuild scanning all model elements, rather than a localized update of just the changed constraint's DOF footprint.

**Failure scenario:** In a large model (e.g. 100k elements) with active sliding contact that reassigns a handful of slave-facet pairs almost every increment, `connectivityHasChanged` is `True` on nearly every increment -- forcing a full O(total elements + total DOFs) rebuild every increment for the whole analysis, even though only a small contact patch actually changed.

alexdummer:
Three more findings from the same review, but on code this PR doesn't touch (so GitHub won't let me attach them as inline diff comments) — flagging for awareness since they directly affect whether the new deformable-surface contact feature works correctly under explicit solvers:

**`edelweissfe/solvers/nonlinearexplicitdynamic.py:160`** — `V = self.theDofManager.constructDofVector()` is never seeded from element/rigid-body `initialVelocity`. `BaseElement.initialVelocity` (`baseelement.py:239`) and `PointMass.initialVelocity` (`pointmass.py:89`) are wired through `discreterigidbodygenerator.py`, but nothing reads `element.initialVelocity` to populate `V` in explicit dynamics.
- *Failure scenario:* A `*rigidBody`/`PointMass` configured with `initialVelocity=[10,0,0]` for an explicit-dynamics impact/drop test starts at rest instead of moving — silently, no warning or error.

**`edelweissfe/solvers/nonlinearexplicitstatic.py:203`** — NEST/NED build the `DofManager` once at step start and never call `constraint.updateConnectivity()`, unlike NIST (`nonlinearimplicitstatic.py:163`). `NodeToDeformableSurfacePenalty` leaves `_nodes`/`_fieldsOnNodes`/`_nDof` empty until `updateConnectivity()` runs.
- *Failure scenario:* Running the new deformable-surface contact under `solver=NEST` or `NED` yields a contact constraint contributing zero DOFs for the whole analysis — no penetration prevention, no error. All existing contact tests use NIST/NISTParallel, so this gap is untested.

**`edelweissfe/rigidbodies/discreterigidbody.py:240`** (pre-existing, unchanged by this PR) — `DiscreteRigidBody` parameterizes rotation as an SO(3) exponential-map pseudo-vector (Rodrigues via `rotationMatrixFromPseudoVector`), while the pre-existing `edelweissfe/constraints/rigidbody.py` parameterizes the same conceptual `'rotation'` `NodeField` DOF as Euler angles (`Rx`/`Ry`/`Rz` composed as `Rz*Ry*Rx`) — two incompatible conventions for the same field name.
- *Failure scenario:* A `'rotation'` `NodeField` value produced by one rigid-body mechanism is not interpretable by the other's kinematics. Any future generic consumer of `nodeFields['rotation']` (a restart/checkpoint path, a generic visualizer, or code mixing the old `rigidbody` constraint with the new `DiscreteRigidBody` in the same model) must special-case which convention produced the data, or silently misinterprets small-angle Euler values as a Rodrigues pseudo-vector (or vice versa).

edelweissfe/generators/surfaceelementgenerator.py
matthiasneuner: Fixed properly rather than just guarded against — turned out to be a bigger issue than a face-numbering label mismatch: this codebase's own `Hexa8`/`Hexa20` elements used a genuinely different local node ordering than Marmot's (confirmed against Marmot's actual C++ shape functions), not just a different face-number convention on the same ordering.

Unified onto Marmot's node ordering in [PR #83](https://github.com/Edelweiss-Numerics/EdelweissFE/pull/83) (cherry-picked here as `a4dcef19`), then re-derived `_FACE_TABLES`/`_MIDSIDE_FACE_TABLES` from Marmot's own `getBoundaryElementIndices` in `9ebc3d9b`. Verified numerically (not assumed) that `boxGen`/`pipeGen` and the one hand-written hex mesh in the test suite need no changes — the old and new conventions are related by a consistent, orientation-preserving axis permutation. Full `edelweiss-only` suite: 0 failures; end-to-end verified that a real Abaqus `*surface, S1` definition now produces the correct facet on a real pure-Python `C3D8`.

edelweissfe/constraints/nodetodeformablesurfacepenalty.py
matthiasneuner: Fixed in `01aea24d`. Rather than a discrete "did the facet change" check, detects the discontinuity via the angle between the previous and newly frozen facet normals (30 degrees separates smooth facet-to-facet drift on an adequately resolved curved surface from an actual crease, with margin on both sides -- justified in the commit message). When exceeded, discards the tangential history instead of rotating it, but keeps the normal traction multiplier (frame-independent, and releasing it would undo converged Uzawa progress for no correctness benefit).

No existing test exercises this path -- every current master surface is flat or smoothly curved -- so this is a fix for a scenario nothing currently tests, not a regression fix. Full suite unaffected (0 failures).

edelweissfe/rigidbodies/discreterigidbody.py
matthiasneuner: This is now moot: the review ran against b715bf6a, which predated a cherry-pick (cc2b0018) of PR #54's own final review-feedback commit that #57's rebase had missed. That commit already restricted `DiscreteRigidBody` to 3D-only via a `WrongDomain` check in the generator (the only construction path for this class), so `getCurrentKinematics` no longer needs its own domainSize guard -- confirmed against the current tip.

edelweissfe/constraints/nodetodeformablesurfacepenalty.py
matthiasneuner: Fixed in `62babeee`: both divisions now use `np.divide(..., out=np.zeros_like(...), where=tributaryAreas > 0)`, reporting 0 pressure/traction for a degenerate zero-area slave node instead of NaN. Verified the guard behaves correctly on a synthetic zero-area case; full suite unaffected (0 failures).

edelweissfe/rigidbodies/discreterigidbody.py
matthiasneuner: Also moot, same root cause as the domainSize comment above: fixed by the cc2b0018 cherry-pick of PR #54's final review-feedback commit. `getVisualizationField` now raises `ValueError` for an unsupported field name instead of silently returning zeros -- confirmed against the current tip.

edelweissfe/constraints/nodetodeformablesurfacepenalty.py
matthiasneuner: Agreed, and deferred deliberately rather than fixed here: a proper fix (a spatial acceleration structure over master facets, mirroring the rigid-body path's VTK locator) is a genuine feature, not a hotfix, and doesn't block correctness of anything currently tested. Tracked as follow-up perf work; not addressed in this PR.

edelweissfe/solvers/nonlinearimplicitstatic.py
matthiasneuner: This is a deliberate, documented tradeoff rather than an oversight -- the comment right above this line explains it mirrors EdelweissMeshfree's `NonlinearQuasistaticSolver`, which already rebuilds its equation system per-increment on the same kind of signal. You're right about the cost in the scenario you describe (near-every-increment full rebuild under active sliding contact in a large model); no action taken here, but flagging it as a known cost of the current design rather than disputing the finding.

matthiasneuner:
Responding to the three findings above (on code this PR doesn't directly touch):

**`nonlinearexplicitdynamic.py:160` (`initialVelocity` never wired into explicit dynamics)** — moot. This entire API (`BaseElement`/`PointMass` `initialVelocity`, end-to-end through the rigid body and generator) was removed as unused, never-wired-up dead code by the `cc2b0018` cherry-pick of PR #54's final review-feedback commit, which #57's original rebase had missed. There's nothing left to wire up.

**`nonlinearexplicitstatic.py:203` (NEST/NED never call `updateConnectivity()`)** — fixed in `eabc66e5`. Rather than attempt to wire live connectivity updates into two more solvers (real scope, not addressed here), both NEST and NED now raise a clear error at `solveStep` entry if any constraint requires it -- converting the silent "contact contributes zero DOFs for the whole analysis" failure mode into an immediate, loud one. `NESTParallel`/`NEDParallel` inherit this unchanged since they subclass without overriding `solveStep`. Contact is not currently supported under explicit dynamics/statics; this fix doesn't change that, it just stops it from failing silently.

**`DiscreteRigidBody` SO(3)/Rodrigues vs. `rigidbody.py`'s Euler-angle `'rotation'` convention mismatch** — left as-is, per your own framing: pre-existing, unchanged by this PR, and affects the *other* rigid-body mechanism just as much. Tracking as a known cross-cutting inconsistency rather than fixing it here; happy to open a follow-up issue for it if useful.

matthiasneuner:
Replying to the [suppressed-comments review](https://github.com/Edelweiss-Numerics/EdelweissFE/pull/57#pullrequestreview-4947214992) (11 findings folded into a collapsed `<details>` block, not individually addressable review threads). Addressed in `4d1a2eeb`..`cc04a833`:

- **`surfaceelementgenerator.py:228`** (non-deterministic facet/tie-break order) — fixed at the root cause: `abqmodelconstructor.py`'s manual `*elSet` construction used `set(els)`, which is what actually made iteration order non-deterministic (boxGen's own sets were already order-preserving). Switched to `dict.fromkeys(els)` to dedupe while preserving first-occurrence order. `4d1a2eeb`
- **`nodetodeformablesurfacepenalty.py:413`** (`penalty` unvalidated) — now rejects non-positive values at construction. `f4b6ea49`
- **`nodetodeformablesurfacepenalty.py:431`** (`tangentPenalty` unvalidated) — now rejects non-positive values when `mu > 0`. `f4b6ea49`
- **`nodetodeformablesurfacepenalty.py:134` (+413/431/524)** (`augmentedLagrange` docstring states the linear-law update unconditionally) — reworded to describe the actual law-dependent update (`acceptLastState()` already did the right thing; only the public option docstring was misleading). `f4b6ea49`
- **`test_mismatched.inp:12`** (contradicts lines 1-4 on `pNspread`) — ran the test to check which claim was true: `pNspread` is ~0.47, matching lines 1-4. The "stays at 1e-3 / recovers uniform pressure" text was simply wrong and has been corrected. `cc04a833`
- **`FrictionHexa20/test.inp:80`, `ALHexa20/test.inp:80`** (comment says drag 0.8, BC is 0.4) — comments corrected to 0.4. `cc04a833`
- **`ALHexa20/test_slideOffEdge.inp:82`** (comment says 0.8, BC is 1.4) — corrected to describe the actual slide-off-the-edge scenario. `cc04a833`
- **`surfaceelementgenerator.py:186` (+227)** (`area / 4` split is only exact for affine quads) — not fixed; this is a real accuracy question worth investigating separately (computing shares from the actual bilinear shape-function/Jacobian integration). Documented as a known limitation directly on `_assignQuadConsistentShares` for future work. `ff52b197`

Two findings are left as accepted trade-offs, no code change:
- **`nodetodeformablesurfacepenalty.py:529`** (closest-facet search is O(slaves × facets), not a true broadphase) — correct; a spatial-index broadphase is a legitimate follow-up but out of scope for this PR.
- **`contactsurfaceelement.py:96`** (geometry-only facets counted by `DofManager`) — the specific "spatial-average flux denominator" mechanism cited doesn't exist anywhere in this codebase's `outputmanagers`/`fields`/`elementresultcollector`, so that part doesn't apply here; the VIJ-sizing/no-op-kernel cost is real but minor (facet elements have no material DOFs to eliminate) and not worth a design change for this PR.

Full suite re-verified after these changes (all `NodeToDeformableSurfaceContact*`/`NodeToDiscreteRigidBodyContact` tests pass; `pre-commit` clean).

PR #59, opened 2026-07-20
refactor: Step managment system

matthiasneuner:
Structural changes

- StepBase is now a real base class holding all shared step implementation (attributes, solve(), the full incrementation API). AdaptiveStep and AdaptiveStepForExplicitSimulations shrank to ~75-line subclasses that only choose their time stepper. A new TimeStepperBase defines the stepper interface; AdaptiveTimeStepper gained changeIncrementSize and SimpleTimeStepper gained preventIncrementIncrease, so any step type now works with any solver — the mid-run AttributeErrors are gone.
- Step types dispatch through a registry in config/steps.py (mirroring config/solvers.py); an unknown type now raises with the available types listed instead of silently skipping the step.
- Step actions register themselves against inputLanguage["step"].modules instead of hardcoding the two step-type names in every one of the 14 files — a new step type picks up all actions automatically. They also import InputLanguage from utils now, not from steps.adaptivestep.
- StepManager.dequeueStep became generateSteps — non-mutating (re-iterable definitions), proper ValueError for duplicate action names, and the actions live in a StepActionCollection that raises on unknown category names instead of defaultdict-swallowing typos.
- The parser's update mechanism (>>dirichlet with an existing name) is generalized: it now probes for a registered update<keyword> instead of the hardcoded adaptive-only action list, so updates work in explicit steps too.

Bugs fixed along the way

- The >>options regression: options actions are keyed by category again, and solver options are routed through new registerOptionsArg/getOptionsOfCategory helpers in stepactions/options.py. All option args register with a None default meaning "unspecified", so consumers receive exactly what the user typed — this also fixes a second latent problem where injected parse-time defaults would have crashed _updateOptions or reset *solver-dataline settings. I verified end-to-end that NIST now logs Updating option extrapolation=linear for an input using >>options. NEST's runge-kutta options are now registered (they were previously unparseable), and Ensight uses the same lookup.
- Arc-length solver: bodeforce → bodyforce (body forces got the wrong reference-load magnitude at step end), and stepActions["geostatics"] iterating keys and calling a nonexistent apply() → stepActions["geostatic"].values() + applyAtIterationStart().
- options.StepAction.__getitem__ missing return.
- indirectcontractioncontrol was triply bit-rotted (wrong computeDDLambda/finishIncrement arities, applyAtStepEnd(U, P), and stale node-field indexing); it now matches the controller interface and resolves dof indices through the current DofManager.

### DIFF
--- a/edelweissfe/config/steps.py
+++ b/edelweissfe/config/steps.py
@@ -25,40 +25,48 @@
 #  The full text of the license can be found in the file LICENSE.md at
 #  the top level directory of EdelweissFE.
 #  ---------------------------------------------------------------------
-# Created on Mon Jan 23 13:05:53 2017
-
-# @author: Matthias Neuner
 """
-Step actions define actions run during a simulation.
-They are defined within a ``*step`` definition, line by line,
-by specifying their ``name`` and a list of ``option=value``, for example
+Steps are defined via the ``*step`` keyword, and the step type is chosen
+via the ``type`` option:
 
 .. code-block:: edelweiss
 
-    *step, solver=mySolver
-    >>dirichlet, name=bottom,   nSet=bottom,   field=displacement, 2=0, 1=0
-    >>dirichlet, name=rightTop, nSet=rightTop, field=displacement, 2=0,
-    >>distributedload, name=dload1, surface=top,  type=pressure, magnitude=10, f(t)=t*2
-    >>distributedload, name=dload2, surface=left, type=pressure, magnitude=50, f(t)=t
+    *step, solver=mySolver, type=adaptive
 """
 
 import importlib
 
+stepLibrary = {
+    "adaptive": ("edelweissfe.steps.adaptivestep", "AdaptiveStep"),
+    "adaptiveForExplicitSimulations": (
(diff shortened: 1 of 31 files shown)


matthiasneuner:
Closing without merge: audited this branch's content against current `next_v26.11` (2026-08-19) while preparing the #68 transplant.

`#59` has exactly 3 commits beyond trunk:
- `38fc7aff` (the step-management overhaul) — already cherry-picked into the #64→#68 chain via the intermediate `feat/step-management-on-amr` branch, and since further evolved through 76 additional input-system commits on top of it. Verified function-by-function that nothing was lost in that evolution: the only two functions present in `38fc7aff` but absent downstream (`addIncrementationOptionsToModule`, `getModuleArgNames`) are pure helpers for the old `Module`/`InputLanguage` grammar system, retired wholesale in favor of typed schema dataclasses — the functionality itself (standard incrementation options, arg validation) is still there, just via a different mechanism.
- 2 commits pinning black's `target-version` to `py312` — already on trunk.

Nothing here is unique or unmerged elsewhere. Retiring rather than rebasing/merging separately. Branch left in place for now; full audit trail in `HANDOFF_PR_STRATEGY.md`.

PR #61, opened 2026-07-20
feat: surface tie constraint

matthiasneuner:


edelweissfe/solvers/base/nonlinearsolverbase.py
```diff
@@ -55,6 +56,12 @@ class NonlinearSolverBase(ABC):
 
     SolverSpecificOptions = {}
 
+    #: The active multi-point-constraint (hanging node / tie) condensation, if any.
+    #: None for solvers that never build one (e.g. the explicit dynamic solvers), so
+    #: dirichlet.applyDirichletToStiffness can tell an MPC-transformed (fresh, disposable)
+    #: system matrix apart from the assembler's own persistent, in-place-updated one.
+    mpcTransformation = None
```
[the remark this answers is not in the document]
matthiasneuner: Fixed in eb34c902.

edelweissfe/constraints/tie.py
```diff
@@ -0,0 +1,289 @@
...
+
+        closestPointFunction = tria3ClosestPoint if self.nDim == 3 else line2ClosestPoint
+        masterFacetCoords = [np.array([n.coordinates for n in el.nodes]) for el in masterFacetElements]
+
+        # Tie MEMBERSHIP (is this node tied at all) is independent of adjust (does a tied node get
+        # snapped) -- these are separate decisions, see the adjust arg's docstring. A single
+        # tolerance, computed once from the whole master surface's characteristic facet size, is
+        # used for every slave node (matching Abaqus' *TIE, which always enforces some tolerance --
+        # explicit or internally computed -- and never ties unconditionally regardless of distance).
+        membershipTolerance = (
```
alexdummer, line 229: **Bug (crash):** An empty or misconfigured `masterSurface` (0 facets) makes `membershipTolerance` `NaN` via `np.mean([])` on `masterFacetCoords`. The inner facet loop then never runs, so `bestDistance` stays `inf` and `bestFacetIdx` stays `None`. The guard at line 243 (`bestDistance > membershipTolerance`) evaluates `inf > nan` = `False`, so the untied branch is skipped and execution falls through to `masterFacetElements[bestFacetIdx]` with `bestFacetIdx=None`, raising an opaque `TypeError` instead of a clear validation error naming the empty master surface.

Root cause: this closest-facet search is adapted from `nodetodeformablesurfacepenalty.py:483`, which guards with `if bestFacet is not None and ...` — that guard was dropped here.

edelweissfe/constraints/tie.py
```diff
@@ -0,0 +1,289 @@
...
+            if bestDistance > membershipTolerance:
+                self.untiedSlaveNodes.append(slaveNode)
+                continue
+
+            # Snapping is the separate, independent decision: a tied node is only snapped if adjust
+            # is requested AND (when given) its distance is also within adjustTolerance -- a node
+            # beyond adjustTolerance stays tied, just not snapped, preserving its gap exactly.
+            withinAdjustTolerance = adjustTolerance is None or bestDistance <= adjustTolerance
+            if adjust and withinAdjustTolerance and bestDistance > 0.0:
+                slaveNode.coordinates[:] = bestWeights @ masterFacetCoords[bestFacetIdx]
```
alexdummer, line 261: **Correctness risk:** `adjust=True` mutates slave node coordinates in place with no ordering enforcement relative to other constraints that may have already cached those nodes' coordinates by value (e.g. a `nodeToDeformableSurfacePenalty` contact constraint on an overlapping node set constructed earlier). If such a constraint's `__init__` copies its reference coordinates before this tie snaps them, the contact silently runs against stale, pre-snap geometry. Only the docstring warns about this; nothing enforces construction order in `abqmodelconstructor.py`.

edelweissfe/solvers/nonlinearimplicitstatic.py
```diff
@@ -566,7 +580,32 @@ def computeBodyForces(
...
+        #   belongs to which (row, col) -- silently misaligned values from the next
+        #   iteration on. Must NOT eliminate.
+        # - MPC transformation active (hanging nodes / ties): K is the freshly computed,
+        #   disposable T^T @ K @ T + C from mpcTransformation.transformSystemMatrix,
+        #   independent of the generator's buffer and discarded after this solve.
+        #   Eliminating here is always safe, and PARDISO's reordering on these more
+        #   poorly conditioned, path-dependent (contact/friction) condensed systems is
+        #   sensitive enough to the extra explicit-zero structural entries to visibly
+        #   drift from the converged reference path if they are kept.
+        if self.mpcTransformation is not None:
```
alexdummer, line 609: **Fragile invariant:** gating `K.eliminate_zeros()` purely on `self.mpcTransformation is not None` uses an unrelated attribute as a proxy for "is `K` the disposable MPC-transformed matrix vs. the assembler's persistent aliased CSR buffer." This exact bug class required a dedicated prior fix (commit `fb4d15f8`) — calling `eliminate_zeros()` on the assembler's aliased buffer desyncs `CSRGenerator`'s fixed assembly map from the compacted indices, silently corrupting values from the next Newton iteration on. Any future code path that sets `mpcTransformation` for an unrelated reason while `K` is still the assembler's own buffer reintroduces this corruption, since the guard isn't tied to `K`'s actual provenance.

edelweissfe/solvers/nonlinearexplicitstatic.py
```diff
@@ -208,6 +208,11 @@ def solveStep(
                     "contact is not currently supported with this solver."
                 )
 
+        if model.multiPointConstraints:
```
alexdummer: **Maintainability/altitude:** the "multi-point constraints unsupported" guard is hand-duplicated per solver (this file, `nonlinearimplicitstaticparallelarclength.py`) instead of being a single check in `NonlinearSolverBase`. A future solver subclass that forgets to copy this guard would silently build an equation system that never condenses the slave DOFs, producing a physically wrong (unconstrained slave) solution with no error at all.

edelweissfe/constraints/tie.py
```diff
@@ -0,0 +1,289 @@
...
+            model.nodeSets[f"{name}_tied"] = NodeSet(f"{name}_tied", tiedNodes)
+        if self.untiedSlaveNodes:
+            model.nodeSets[f"{name}_untied"] = NodeSet(f"{name}_untied", self.untiedSlaveNodes)
+
+    def getMultiPointConstraints(self, dofManager) -> list[tuple[int, list[tuple[int, float]]]]:
+        fieldVariableIndices = dofManager.idcsOfFieldVariablesInDofVector
+
+        records = []
+        for slaveNode, masterNodes, weights in self.tiedRecords:
+            slaveDofIndices = fieldVariableIndices[slaveNode.fields["displacement"]]
```
alexdummer: **Robustness:** `fieldVariableIndices[node.fields["displacement"]]` is a direct dict lookup with no checked fallback. If `slaveSurface`/`masterSurface` reference facet elements built from nodes that aren't otherwise part of any section-assigned element (orphaned/geometry-only nodes), this raises a bare `KeyError` deep inside the solver's MPC-transformation build step rather than a clear constraint-construction-time error naming the offending node/surface.

edelweissfe/constraints/tie.py
```diff
@@ -0,0 +1,289 @@
...
+        # Exposed as ordinary node sets so the tied/untied split is directly inspectable -- e.g. via
+        # *fieldOutput, or for free in Ensight, which automatically creates a part for every node
+        # set in the model (edelweissfe/outputmanagers/ensight.py's _createGeometryParts), no
+        # fieldOutput required just to see where these nodes are. A set with no members is left
+        # unpublished rather than published empty: Ensight's export is unconditional over every
+        # node set, so a tie whose untied side is (as is typical) always empty would otherwise get
+        # its own empty, useless part in every export.
+        tiedNodes = [record[0] for record in self.tiedRecords]
+        if tiedNodes:
+            model.nodeSets[f"{name}_tied"] = NodeSet(f"{name}_tied", tiedNodes)
```
alexdummer: **Robustness:** `model.nodeSets[f"{name}_tied"]` / `[f"{name}_untied"]` are assigned with no collision check against pre-existing node set names. If a constraint's name makes `{name}_tied`/`{name}_untied` collide with a node set already created by a generator or another constraint, this silently overwrites the earlier set; later `*fieldOutput`/`*dirichlet` references to the old name would silently see the tie's nodes instead.

edelweissfe/solvers/nonlinearexplicitdynamic.py
```diff
@@ -414,9 +431,15 @@ def solveIncrement(
...
+        # fold the forces acting on slave DOFs onto their masters (action-reaction through the
+        # rigid interpolation link); done here so the Dirichlet handling at the start of the next
+        # increment operates on the already-folded vector
+        if self.mpcTransformation is not None:
+            P[:] = self.mpcTransformation.foldExplicitForce(P)
+
         if timeStep.number % self.options["output-frequency"] == 0:
             Wint = psi
-            Wkin = 0.5 * np.sum(1 / Minv * V**2)
+            Wkin = 0.5 * np.sum(self._lumpedMass * V**2)
```
alexdummer: **Diagnostic accuracy (not physics-affecting):** kinetic energy is computed as `self._lumpedMass * V**2` after `foldLumpedMass()` has zeroed each tied slave DOF's own mass, so a tied slave's actual velocity contributes nothing to the reported energy. On a non-matching or bending/rotating tie interface where a slave node's interpolated velocity differs from any single master's velocity, the periodic "Kinetic energy" journal diagnostic under-reports true kinetic energy of the tied region. Time integration itself is unaffected — only this printout.

alexdummer (commented):
Automated code review (via `/code-review --comment`). Findings below as inline comments, plus three more on unchanged code the diff doesn't touch (GitHub can't anchor inline comments there):

- **`edelweissfe/models/femodel.py:268`** — `advanceToTime()` calls `acceptLastState()` on every entry of `model.constraints` but never on `model.multiPointConstraints`, excluding ties from the per-increment state-acceptance lifecycle. Harmless today since `Tie` freezes its records at construction with no trial/converged state, but any future `MultiPointConstraintBase` subclass carrying per-increment history (e.g. adaptive re-projection) would silently leak transient state across increments — `MultiPointConstraintBase` defines no equivalent lifecycle hook either.
- **`edelweissfe/models/femodel.py:355`** — the model summary printer reports `len(model.constraints)` but never mentions `model.multiPointConstraints`. A model with only a tie constraint (no ordinary constraints) shows no "constraints:" line at all in the startup journal, misleading anyone checking the log into thinking no constraints are defined even though a tie is active.
- **`edelweissfe/constraints/nodetodeformablesurfacepenalty.py:483`** — `tie.py`'s closest-facet search duplicates this file's brute-force loop but drops the `bestFacet is not None` safety guard present here — the root cause of the empty-`masterSurface` crash flagged inline on `tie.py:220`. A shared `findClosestFacet` helper (e.g. in `facetcontactgeometry.py`) would fix that crash and prevent the two implementations from diverging further as either evolves (e.g. spatial acceleration structures).

edelweissfe/constraints/tie.py
matthiasneuner: Fixed in e12da74e. Added explicit non-empty validation for both `masterSurface` and `slaveSurface` with descriptive `ValueError`s naming the offending surface, plus an explicit `bestFacetIdx is None` guard in the search loop.

edelweissfe/constraints/tie.py
matthiasneuner: Fixed in e12da74e. `abqmodelconstructor.py` now partitions constraint definitions and instantiates `MultiPointConstraintBase` subclasses (including ties with `adjust=True`) before ordinary constraints, ensuring geometric adjustments take effect before reference coordinates are cached by value.

edelweissfe/solvers/nonlinearimplicitstatic.py
matthiasneuner: Acknowledged. We are keeping this as-is for now and will overhaul the CSR buffer compaction lifecycle and matrix provenance tracking in a dedicated follow-up issue once this PR is merged.

edelweissfe/solvers/nonlinearexplicitstatic.py
matthiasneuner: Fixed in e12da74e. Centralized capability validation via `supportsMPC` and `validateModelCapabilities(model)` on `NonlinearSolverBase`, replacing hand-duplicated checks across solvers.

edelweissfe/constraints/tie.py
matthiasneuner: Fixed in e12da74e. Added checked lookup in `getMultiPointConstraints()` that raises a descriptive `KeyError` naming the constraint and offending node label if displacement DOFs are missing or unregistered.

edelweissfe/constraints/tie.py
matthiasneuner: Fixed in e12da74e. `tie.py` now checks if `{name}_tied` or `{name}_untied` already exists in `model.nodeSets` and raises a `ValueError` rather than silently overwriting it.

edelweissfe/solvers/nonlinearexplicitdynamic.py
matthiasneuner: Fixed in e12da74e. NED now retains `_rawLumpedMass` before `foldLumpedMass()` and uses it for the periodic kinetic energy diagnostic, accurately accounting for slave velocities across bending, rotation, and mismatched interfaces.

matthiasneuner:
All review comments and findings have been addressed in commit `e12da74e`:

- **Master/Slave surface validation & search guards:** Added explicit validation in [`tie.py`](file:///home/matthias/constitutive_modeling/next_v2611/EdelweissFE/edelweissfe/constraints/tie.py) ensuring surfaces contain facet elements with descriptive `ValueError`s, plus a `bestFacetIdx is None` check in the closest-point search loop.
- **In-place coordinate adjustment ordering:** [`abqmodelconstructor.py`](file:///home/matthias/constitutive_modeling/next_v2611/EdelweissFE/edelweissfe/generators/abqmodelconstructor.py) now instantiates multi-point constraints before ordinary constraints so that `adjust=True` geometric snapping takes effect before other constraints cache coordinates by value.
- **Centralized solver capability check:** Added `supportsMPC` and `validateModelCapabilities(model)` on [`NonlinearSolverBase`](file:///home/matthias/constitutive_modeling/next_v2611/EdelweissFE/edelweissfe/solvers/base/nonlinearsolverbase.py) to replace hand-duplicated guards across solver subclasses.
- **Checked DOF resolution:** `getMultiPointConstraints()` now validates displacement field existence and raises clear, actionable `KeyError`s naming the constraint and offending node label.
- **NodeSet collision prevention:** `tie.py` checks whether `{name}_tied` / `{name}_untied` exists in `model.nodeSets` and raises `ValueError` rather than silently overwriting.
- **NED kinetic energy diagnostic:** Retains `_rawLumpedMass` prior to `foldLumpedMass()` for exact kinetic energy accounting in journal printouts across mismatched/bending interfaces.
- **Lifecycle & summary integration:** Added `acceptLastState()` hook on [`MultiPointConstraintBase`](file:///home/matthias/constitutive_modeling/next_v2611/EdelweissFE/edelweissfe/constraints/base/multipointconstraintbase.py), invoked in [`FEModel.advanceToTime()`](file:///home/matthias/constitutive_modeling/next_v2611/EdelweissFE/edelweissfe/models/femodel.py), and included multi-point constraints in the pretty model summary startup printout.
- **CSR zero-elimination guard (`K.eliminate_zeros()`):** Kept as-is for this PR per plan; a dedicated follow-up issue will overhaul the sparse matrix compaction and CSR lifecycle once merged.
- **Unit tests:** Added test suite in [`test_tie.py`](file:///home/matthias/constitutive_modeling/next_v2611/EdelweissFE/edelweissfe/constraints/test_tie.py) covering all edge cases and validations.

PR #64, opened 2026-07-25
feat: adaptive mesh refinement with hanging nodes

matthiasneuner:


edelweissfe/generators/discreterigidbodygenerator.py
```diff
@@ -0,0 +1,421 @@
...
+                "PointMass only supports axis-aligned rotary inertia. Results will be approximate "
+                "unless the body's principal axes are aligned with the global axes.",
+                "discreteRigidBody",
+                0,
+            )
+
+        if mass is None:
+            mass = massProperties.mass
+        if inertia is None:
+            inertia = list(np.diag(massProperties.inertia))
```
alexdummer, line 244: **CONFIRMED — missing parallel-axis correction for rigid-body inertia**

`inertia = list(np.diag(massProperties.inertia))` is the inertia about `massProperties.centerOfMass` (see `polyhedronmassproperties.py`), but `rp_coordinate` a few lines below is only defaulted to the center of mass when the user didn't supply one — a user-supplied `rpCoordinate` (e.g. a hinge/pivot point) is left untouched. The COM-based inertia tensor is then handed to `DiscreteRigidBody`/`PointMass`, which writes it straight into the rotational-DOF slots at the RP node (`pointmass.py: computeLumpedInertia`), i.e. treated as if it were already about the RP.

Whenever a user supplies both `density` and an `rpCoordinate` different from the true center of mass, the `m*(d·d·I − d⊗d)` parallel-axis offset for `d = rpCoordinate − centerOfMass` is silently omitted — angular acceleration under torque and rotational-oscillation frequency come out wrong with no warning.

edelweissfe/modelmodifiers/adaptivity/hadaptivity.py
```diff
@@ -0,0 +1,568 @@
...
+                marked_elements.update(elements)
+
+        # dynamic markers (not initialOnly) need a non-zero state to evaluate safely
+        dynamic_markers = [m for m in self.markers if not m.initialOnly]
+
+        stateMagnitude = max(
+            (float(np.abs(np.asarray(nf["U"])).max()) for nf in model.nodeFields.values() if "U" in nf),
+            default=0.0,
+        )
+        if stateMagnitude >= 1e-12:
```
alexdummer: **CONFIRMED — dynamic AMR markers gated on displacement magnitude regardless of what they evaluate**

`stateMagnitude` is computed only from `nf["U"]` across node fields, and `if stateMagnitude >= 1e-12:` gates out *all* non-`initialOnly` markers when it's below that threshold. At t=0, or in any quasi-static run where displacements stay numerically tiny (stiff structures, thermal/eigenstrain-driven stress with near-zero displacement, small-unit scaling), this silently blocks refinement even for a marker that evaluates stress/strain/some other field entirely unrelated to `U` magnitude. No error, no log — refinement just never triggers.

edelweissfe/solvers/nonlinearexplicitstatic.py
```diff
@@ -174,7 +174,9 @@ def __init__(self, jobInfo, journal, **kwargs):
         self.journal = journal
 
         self.options = self.SolverSpecificOptions.copy()
-        self._updateOptions(kwargs, journal)
+        # the datalines of the *solver keyword belong exclusively to this solver, so unknown entries
+        # are user typos and must not be swallowed
+        self._updateOptions(kwargs, journal, strict=True)
 
     def solveStep(
```
alexdummer, line 185: **CONFIRMED (with one correction from initial review) — explicit solvers never poll `model.modelModifiers` or call `constraint.updateConnectivity()`**

`NonlinearImplicitStatic.solveStep` was extended (nonlinearimplicitstatic.py:207/210) to call `modifier.updateModel(...)` for every `model.modelModifiers` entry and `constraint.updateConnectivity(model)` for every constraint each increment, driving AMR refinement and hanging-node/contact reconciliation. Neither `NEST.solveStep` here nor `NED.solveStep` (nonlinearexplicitdynamic.py:105) does either — confirmed via repo-wide grep, both symbols only appear in `nonlinearimplicitstatic.py`.

A user configuring an `hadaptivity` model modifier with `solver=NEST` or `NED` gets no error and no refinement — the mesh silently never adapts. (Correction from the initial pass: NED does not reject pre-existing MPCs the way NEST does, but the core `modelModifiers`/`updateConnectivity` gap holds for both.) Worth at least a startup warning/`NotImplementedError` analogous to the existing MPC guard, if explicit AMR support isn't planned.

edelweissfe/utils/fieldoutput.py
```diff
@@ -173,6 +176,17 @@ def getResultHistory(
             raise Exception(
                 "fieldOutput {:} does not save any history; please define it with saveHistory=True!".format(self.name)
             )
+
+        if self.result:
+            firstShape = np.shape(self.result[0])
+            if any(np.shape(r) != firstShape for r in self.result):
+                raise Exception(
+                    "fieldOutput {:} has a per-node/per-element history with varying shapes across "
```
alexdummer, line 184: **CONFIRMED — new shape-validation exception breaks two pre-existing, untouched callers under AMR**

`getResultHistory()` now raises when stored per-increment results have varying shapes (explicitly to handle AMR-driven mesh changes, per the docstring above). But `meshplot.py:342/347` (`xyJob["y"/"x"].getResultHistory()`) and `fractureenergyintegrator.py:147` (`np.trapz(self.fpF.getResultHistory(), x=self.fpU.getResultHistory())`) both call it with no surrounding try/except inside `finalizeJob()` — and neither file is touched by this PR. `_resultTableForExport` (used by `writeLastResult()`) already has a graceful fallback for exactly this case; these two callers don't. Running a job with AMR enabled plus a meshplot `xyJob` or fracture-energy integrator on a `saveHistory=True` field output whose shape changes across increments will now crash `finalizeJob()` where it previously didn't.

edelweissfe/solvers/nonlinearimplicitstaticparallelarclength.py
```diff
@@ -70,6 +70,11 @@ def solveStep(
         fieldOutputController: FieldOutputController,
         outputmanagers: dict[str, OutputManagerBase],
     ):
+        if model.multiPointConstraints:
```
alexdummer: **CONFIRMED — arc-length solver silently ignores MPCs added mid-step by AMR**

This guard rejects `model.multiPointConstraints` only once, at `solveStep` entry. If the step starts with none, it passes; h-adaptivity can add hanging-node entries to `model.multiPointConstraints` later in the same step via the inherited `NIST.solveStep` model-change/rebuild loop. Once that happens, `NISTPArcLength.solveIncrement`'s custom indirect-control branch (active whenever `self.arcLengthController is not None`) assembles `K_0/K_f/R_0/R_f` without ever referencing `self.mpcTransformation` — unlike the base `NIST.solveIncrement`, which condenses via `mpcTransformation.transformResidual`/`transformSystemMatrix`. Nothing restricts pairing an `hadaptivity` model modifier with this solver, so the hanging-node constraint ends up silently unenforced — a wrong solution with no error raised.

edelweissfe/generators/discreterigidbodygenerator.py
```diff
@@ -0,0 +1,421 @@
...
+    model.nodes[referencePoint.label] = referencePoint
+
+    rpNodeSetName = f"{name}_rp"
+    model.nodeSets[rpNodeSetName] = NodeSet(rpNodeSetName, [referencePoint])
+
+    if "all" in model.nodeSets:
+        allNodes = list(model.nodeSets["all"])
+        allNodes.extend(rigidNodes)
+        allNodes.append(referencePoint)
+        model.nodeSets["all"] = NodeSet("all", allNodes)
```
alexdummer, line 279: **PLAUSIBLE — `"all"` NodeSet rebound instead of mutated in place, inconsistent with this PR's own convention**

This PR introduces `OrderedSet.replaceMembers` (`orderedset.py:88`) specifically so consumers holding a reference to `model.nodeSets["all"]` keep seeing membership updates in place — `hadaptivity.py:520/527/539` and `surfaceelementgenerator.py:315/336` both use it for this reason. This generator instead does `model.nodeSets["all"] = NodeSet("all", allNodes)`, creating a brand-new object. Anything that captured the old `"all"` NodeSet by reference before this generator runs would silently stop seeing the newly added rigid-body surface/RP nodes.

Note on scope: I couldn't find a currently-existing consumer that caches this reference early enough to actually go stale (everything found does a fresh dict lookup at point of use), so this reads as a latent robustness gap / inconsistency with the established convention rather than a demonstrated live bug today. Worth using `replaceMembers` here too for consistency, since a future consumer following the new convention could rely on it.

edelweissfe/outputmanagers/ensight.py
```diff
@@ -988,7 +1117,26 @@ def finalizeFailedIncrement(self, **kwargs):
 
     def writeOutput(self, model: FEModel):
         self.timeAtLastOutput = model.time
-        self.ensightCase.setCurrentTime(self.transientTAndFSetNumber, model.time)
+
+        # rebuild parts + variable jobs if the mesh changed (AMR), so geometry and variables match
+        signature = (len(model.elements), len(model.nodes))
```
alexdummer, line 1126: **PLAUSIBLE — AMR mesh-change detection via count-only signature can alias, but the current refine-only AMR can't yet trigger it**

`signature = (len(model.elements), len(model.nodes))` is the sole invalidation signal for rebuilding Ensight's geometry parts and per-element/per-node index mappings after a mesh change — there's no version counter or other check. If a future/simultaneous refine+coarsen preserved both totals (or any two distinct topologies produced the same pair), `mesh_changed` would stay `False` and `_rebuildForMeshChange()` would be skipped, silently mislabeling subsequent output against stale index mappings.

However: this AMR implementation is refine-only today (`AdaptiveMesh.refine()` only subdivides — no `coarsen`/`derefine`/`merge` exists anywhere in `refinement.py`/`marking.py`/`hadaptivity.py`), so element/node counts can only grow, never alias in practice yet. Flagging as a fragile pattern worth hardening (e.g. reuse the model's own topology-version counter, as other observers do) before coarsening is ever added, rather than a live bug today.

edelweissfe/sets/elementset.py
```diff
@@ -42,11 +42,11 @@
 else:
     MarmotMaterialWrappingElement = None
 
-from edelweissfe.sets.orderedset import ImmutableOrderedSet
+from edelweissfe.sets.orderedset import OrderedSet
 from edelweissfe.utils.meshtools import extractNodesFromElementSet
 
 
-class ElementSet(ImmutableOrderedSet):
+class ElementSet(OrderedSet):
```
alexdummer, line 49: **CONFIRMED — `ElementSet`/`NodeSet` lost their fail-fast immutability guard**

`ElementSet` (here) and `NodeSet` (`nodeset.py:34`) both switched their base class from `ImmutableOrderedSet` to the new mutable `OrderedSet`, to support the new in-place `replaceMembers()` AMR needs. That also removes the `TypeError` that previously fired on any `.add()`/`__setitem__()`/`__ior__()` against what used to be an immutable set.

Outside the AMR paths that intentionally call `replaceMembers()`, any existing or future code (constraints, generators, output managers, user scripts) that accidentally mutates an element/node set via the now-permitted `.add()`/item-assignment/`|=` will silently succeed instead of failing fast. Since many objects (surfaces, constraints, output-manager part mappings) hold direct references to the same set instance, an unintended mutation from unrelated code could now silently corrupt shared mesh topology instead of raising immediately as before. Worth considering a narrower API on `OrderedSet` (e.g. keep `replaceMembers` as the only mutator, still reject `.add`/`__setitem__`/`__ior__`) rather than dropping the guard wholesale.

edelweissfe/outputmanagers/ensight.py
```diff
@@ -1028,15 +1176,18 @@ def writeOutput(self, model: FEModel):
...
-            self.intermediateSaveIntervalCounter += 1
+        # register the current time for variables
+        self.ensightCase.setCurrentTime(self.transientVariableTAndFSetNumber, model.time)
+
+        # Rewrite the small .case every step so the on-disk step count always matches the committed
+        # data even if the job aborts before finalizeJob() (the L-panel dies on GCDP return-mapping).
+        # A trailing partial geometry step written just before an abort is simply not counted here, so
+        # the reader never overruns it. The .case is tiny text and the data files are already appended
+        # every step, so there is no benefit to gating this behind intermediateSaveInterval.
+        self.ensightCase.finalize(replaceTimeValuesByEnumeration=False, closeFileHandes=False)
```
alexdummer, line 1194: **CONFIRMED — `intermediateSaveInterval` silently has no effect any more; `.case` rewrite is now O(N²) across a run**

The old gate (`if self.intermediateSaveInterval: if self.intermediateSaveIntervalCounter >= ...: finalize(...)`) was replaced with an unconditional `finalize()` call every `finalizeIncrement()`, per the comment above ("no benefit to gating this behind intermediateSaveInterval"). `intermediateSaveInterval` is still parsed and stored, `intermediateSaveIntervalCounter` is still initialized, but neither is used anywhere else — confirmed via grep. `finalize()` re-loops over every previously recorded `timeSet.timeValues` entry and rewrites the whole `.case` file from scratch each call, so calling it unconditionally every increment turns a linear-per-write op into an O(N²) total cost across a run.

This does look like a deliberate, documented crash-safety tradeoff rather than an accidental regression — but it silently removes a user-facing knob (`intermediateSaveInterval=N` in the input file) with no warning that it's now ignored, and could meaningfully slow down long runs with many increments/parts. Worth at least logging a one-time notice when a job sets `intermediateSaveInterval` that it's currently a no-op, so users aren't debugging a silent perf regression.

alexdummer (commented):
Automated review pass (correctness-focused, recall-biased) over the AMR/hanging-node changes. 10 findings, ranked roughly by severity; each was independently re-verified against this diff before posting. Two (marked PLAUSIBLE) are real gaps in the invariants this PR establishes but no live trigger was found yet — flagging as fragility rather than a confirmed bug. 9 are posted as inline comments below; one (`edelweissfe/constraints/linearizedrigidbody.py`) isn't touched by this PR's diff so GitHub won't let me anchor a line comment there — noting it here instead:

**CONFIRMED — `LinearizedRigidBody` has a stale node snapshot after AMR (missing `updateConnectivity`/`MeshDependent`)**

`linearizedrigidbody.py:159/162`: `self.slaveNodes = [node for node in slaveNodeSet if not node == self.rp]` / `self._nodes = self.slaveNodes + [self.rp]` is a plain-list snapshot built once in `__init__` with no re-fetch. Unlike its sibling constraints (`NodeToRigidSurfacePenalty`, `NodeToDiscreteRigidBodyPenalty`, `NodeToDeformableSurfacePenalty`), this class subclasses only `ConstraintBase` — no `MeshDependent`, no `updateConnectivity` override — so it falls back to the default no-op documented as correct only "for every constraint whose DOF footprint is fixed at construction." `amrtransparencyprobe.py` documents the exact contract this violates: NodeSets are mutated in place by AMR, and any constraint caching a plain-list snapshot from such a set (exactly what this does) will silently miss newly added nodes after a refinement event.

edelweissfe/modelmodifiers/adaptivity/hadaptivity.py
matthiasneuner: Fixed in 1385ecba: replaced the `stateMagnitude >= 1e-12` gate with `not self._isFirstCall`. That correctly captures "has at least one increment actually converged" regardless of which field a marker evaluates, instead of using displacement magnitude as a (wrong) proxy for it. The original intent (per the introducing commit's comment) was only ever to skip the very first pre-solve call, where model fields still hold the initial condition -- `self._isFirstCall` already tracked exactly that.

edelweissfe/solvers/nonlinearexplicitstatic.py
matthiasneuner: Fixed in 1385ecba: added `supportsModelModifiers` alongside the existing `supportsMPC` pattern -- defaults to `False` on the base class, `NIST` sets it `True` (its `solveStep` loop is the only one that polls `model.modelModifiers`), and `NEST`/`NED` now raise `NotImplementedError` via `validateModelCapabilities` if an `hadaptivity` modifier is present. Note `NEST` subclasses `NIST`, so it needed an explicit `False` override (mirroring how it already overrides `supportsMPC`) rather than just relying on the base default.

edelweissfe/generators/discreterigidbodygenerator.py
matthiasneuner: Agreed this is a real bug, but it's pre-existing in `next_v26.11`: this file was taken wholesale in the merge (every commit touching it predates the AMR fork and none are AMR-specific), so it's outside this PR's diff in substance even though it shows up in the merge. Filing it as a separate issue against mainline rather than fixing it here, to keep this PR's scope to the AMR feature itself.

edelweissfe/utils/fieldoutput.py
matthiasneuner: Not fixing in this PR -- it's a real gap (those two callers can crash under AMR + varying-shape result histories) but it's narrow (only hit when AMR is combined with a meshplot `xyJob` or the fracture-energy integrator specifically) and touching `meshplot.py`/`fractureenergyintegrator.py` is outside this PR's actual scope. Tracking it as a follow-up: both callers should get the same graceful-shape-fallback treatment `_resultTableForExport` already has.

edelweissfe/solvers/nonlinearimplicitstaticparallelarclength.py
matthiasneuner: Verified this is already handled, no code change needed: `buildMPCTransformation` (nonlinearsolverbase.py) is called on every DofManager rebuild inside the shared `NIST.solveStep` per-increment loop -- including for `NISTPArcLength`, which inherits that loop via `NISTParallel` -- and it unconditionally raises `NotImplementedError` when `model.multiPointConstraints` is non-empty and `supportsMPC` is `False` (which `NISTPArcLength` sets). That check runs *before* `solveIncrement` is ever dispatched, so the moment h-adaptivity adds a hanging-node MPC mid-step and the DofManager gets rebuilt, this raises loudly rather than silently assembling with the custom indirect-control branch. Traced the call order in nonlinearimplicitstatic.py to confirm (rebuild+buildMPCTransformation at line ~261, solveIncrement dispatch at line ~359, same loop iteration).

edelweissfe/generators/discreterigidbodygenerator.py
matthiasneuner: Agreed on the inconsistency, but per your own note there's no demonstrated live consumer that goes stale from this today, and it's in code taken wholesale from `next_v26.11` (pre-existing, not AMR-specific). Leaving as-is for this PR; worth revisiting if/when a consumer actually starts caching `model.nodeSets["all"]` early.

edelweissfe/outputmanagers/ensight.py
matthiasneuner: Agreed this is fragile, but per your own note it can't alias while AMR is refine-only (no coarsen/derefine exists anywhere yet). Deferring the hardening (e.g. switching to the model's topology-version counter) until coarsening is actually on the table, to keep this PR scoped to what's shipping now.

edelweissfe/sets/elementset.py
matthiasneuner: Agreed this is a real robustness regression in the abstract, but no demonstrated live bug today, and a proper fix (a narrower mutable-but-guarded `OrderedSet` API that keeps `replaceMembers` as the sole mutator while still rejecting `.add`/`__setitem__`/`__ior__`) is more design work than this PR's scope justifies right now. Tracking as a follow-up rather than fixing inline.

edelweissfe/outputmanagers/ensight.py
matthiasneuner: Confirmed as described -- this is the deliberate crash-safety tradeoff we made during the merge (matches your own classification), not an accidental regression. Not adding the one-time notice in this PR to keep the diff focused, but agree it's a good idea; noting it as a follow-up so users setting `intermediateSaveInterval` aren't left debugging a silent no-op.

Issue #72, opened 2026-07-28
Investigate CSR reuse-vs-recreate tradeoff: eliminate_zeros() is unsafe on updateInPlace's aliased matrix, and structurally-zero coupling blocks aren't unique to AMR/MPC


matthiasneuner:
## Background

While tracking down an AMR test regression (`feat/amr-hanging-nodes`), we found that
`CSRGenerator.updateInPlace` (`edelweissfe/numerics/csrgeneratorv2.pyx`) and
`csr_matrix.eliminate_zeros()` are fundamentally incompatible when combined:

- `updateInPlace` returns the generator's own persistent CSR matrix *by reference* and
  rewrites its `data` buffer in place every call, using a gather/scatter map computed
  once, at construction time, from the full COO pattern.
- `eliminate_zeros()` compacts that same matrix's `data`/`indices`/`indptr` arrays
  in place (via `scipy.sparse._sputils._prune_array`, which returns either a view or a
  copy of the original buffer depending on how much shrank).
- Either way, after the first call, the generator's fixed assembly map and the (now
  compacted) `indices`/`indptr` disagree about which stored slot belongs to which
  `(row, col)`. Every subsequent `updateInPlace` call then writes numerically correct
  *values* into the *wrong slots* — a silent correctness bug, not a crash (confirmed
  with an isolated repro: a small dense 3x3 pattern with one Dirichlet-style row
  zeroed showed data misaligned by exactly the number of eliminated entries).

The immediate trigger was `edelweissfe/solvers/base/dirichlet.pyx` briefly calling
`eliminate_zeros()` after zeroing Dirichlet rows on a matrix that, for most solves, *is*
the aliased generator matrix. The fix landed (see `NIST.applyDirichletK` in
`edelweissfe/solvers/nonlinearimplicitstatic.py`) only calls `eliminate_zeros()` when
`self.mpcTransformation` is active, because in that case `K` is a fresh, disposable
`T^T @ K @ T + C` matrix, not the aliased one — safe by construction, not by policy.

## Why this needs more than a local fix

Dirichlet + MPC condensation is only one source of structurally-present-but-zero
entries. The same shape of problem exists wherever a coupling term is part of the
*declared* sparsity pattern (from element/constraint COO contributions) but is only
*numerically* nonzero conditionally — e.g.:

- Gradient-enhanced / nonlocal-damage coupling (`nonlocal damage` field) in elements
  that are still in the elastic regime.
[...]

PR #74, opened 2026-07-31
fix: macOS/arm64 support

mitrlk:
Makes EdelweissFE installable and runnable on macOS/arm64 (Apple Silicon).

1.     conda_requirements.txt pins mkl, mkl-include and intel-openmp, which are x86-only, so environment creation fails outright on osx-arm64. Adds conda_requirements_osx-arm64.txt.

2.     getLinSolverByName("pardiso") hard-imports the MKL backend. It now falls back to MUMPS if available, otherwise SciPy's SuperLU, with a warning — so pardiso-configured input files and the test suite run unchanged. Both fallback paths are exercised.

3.     analyticalfields/fromvtk.py imported pyvista at module scope, but inputfileparser imports that module unconditionally to register its input language, and vtk/pyvista have no free-threaded wheels on osx-arm64 — so every test aborted on import. Made lazy.

4.     linsolve/mumps/mumps.py: b may be a DofVector, whose copy() drops entitiesInDofVector, so np.reshape(b.copy(), ...)[:, i] raises AttributeError. Work on a plain ndarray.

Test suite on Apple M4: 79 passed, 48 skipped, 0 failed.

### DIFF
--- a/.github/workflows/run_tests_with_marmot.yml
+++ b/.github/workflows/run_tests_with_marmot.yml
@@ -18,17 +18,17 @@ jobs:
           activate-environment: anaconda-client-env
 
       - name: Install conda packages
-        shell: bash -l {0}
+        shell: bash -leo pipefail {0}
         run: |
           mamba install --file conda_requirements.txt
 
       - name: Install pip packages
-        shell: bash -l {0}
+        shell: bash -leo pipefail {0}
         run: |
           pip install -r pip_requirements.txt
 
       - name: Install Eigen
-        shell: bash -l {0}
+        shell: bash -leo pipefail {0}
         run: |
           cd ..
           git clone --branch 3.4.0 https://gitlab.com/libeigen/eigen.git
@@ -40,7 +40,7 @@ jobs:
           cd ../..
 
       - name: Install Autodiff
-        shell: bash -l {0}
+        shell: bash -leo pipefail {0}
         run: |
           git clone --branch v1.1.0 https://github.com/autodiff/autodiff.git
           cd autodiff
@@ -56,7 +56,7 @@ jobs:
           cd ../..
 
       - name: Install Fastor
-        shell: bash -l {0}
+        shell: bash -leo pipefail {0}
         run: |
           git clone https://github.com/romeric/Fastor.git
           cd Fastor
@@ -67,7 +67,7 @@ jobs:
           cd ../..
 
       - name: Install AMGCL
-        shell: bash -l {0}
+        shell: bash -leo pipefail {0}
         run: |
           git clone --branch 1.4.7 --depth 1 https://github.com/ddemidov/amgcl.git
           cd amgcl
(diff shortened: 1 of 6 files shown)


alexdummer:
@mitrlk Please have a look at the new next_v26.11 because there is now more dependence on pyvista. 

One more question on lazy-importing pyvista: Is it not possible to install pyvista via pip after conda requirements have been installed?

Additionally, a macos/arm pipeline would make sense here too, such that we make sure to not break support in the future.

mitrlk:
Thanks @alexdummer , rebased onto next_v26.11, conflicts gone. 

Sorry for the longer comment: 

Three of my four changes were no longer needed since you fixed the same things upstream, so I dropped them. What remains is a separate conda_requirements_osx-arm64.txt, with the normal one untouched. It omits mkl, mkl-include and intel-openmp, which are Intel-only and cannot be installed on arm64 at all. It also omits python-freethreading, which answers your pyvista question: pyvista installs fine but vtk does not, since VTK only publishes free-threaded builds for Linux. Dropping it lets vtk install normally. Happy to invert that if you would rather keep the same Python build everywhere.

The macOS job also found a bug, fixed here: csrgeneratorv2.pyx declared mapStarts as long while _csrcore.h declares const int64_t*. On Linux those are the same type, on macOS int64_t is long long, so the extension silently failed to build and every test then skipped. It is a platform difference rather than a compiler one; GCC and clang agree with each other on both systems.

One thing is open, and it is why I would not call this complete yet. The job gets 8 passed and 156 skipped, 112 of them on the missing pardiso module. PARDISO is part of MKL, which is x86 only, so it is unavailable on any arm64 machine. getDefaultLinSolver() is written to fall back to superlu, but it can never run: SolverSpecificOptions sets "linsolver": "pardiso", so "linsolver" in self.options is always true.

I tried making that branch reachable, and superlu is the wrong fallback: every test then fails on "Refusing to call sum_duplicates() on a CSRGenerator.updateInPlace() matrix", which is issue #72. Other solvers do work though. I built the full stack locally on arm64 with mumps, confirmed it reproduces the CI result, and pointed the default elsewhere. The edelweiss-only suite goes from 4 passed and 44 skipped to 44 passed, 4 skipped, 0 failed. The testfiles/marmot suite gives 61 passed, 47 skipped, 8 failed, with no sum_duplicates errors anywhere. Those 8 are all adaptive refinement tests, described below, and are not solver-related. So a working fallback exists without needing #72 resolved, but choosing which one changes default behaviour on all platforms, so I did not want to decide that myself. Happy to implement whichever you prefer, or to leave the solver question separate and take this PR as the build and CI groundwork. Additional note: the same "linsolver": "pardiso" default appears in nonlinearexplicitstatic.py too, so a fix would need to cover both solvers.

Lastly, those 8 adaptive refinement failures. They are not new breakage: those tests have never run on arm64, because they were skipping on the missing pardiso module, so once a working solver is in place they execute for the first time and fail. It is not the linear solver either, I get identical failures with gmres. Run on its own, AMR_DynamicRefinement refines to the expected 279 dofs and converges, but the result differs from U.ref by 5.9e-05 against a reference maximum of 5.0e-04, which is too large to be floating-point noise. Run as part of the full suite the same test instead refines to 423 dofs and fails on a shape mismatch, which looks like state persisting between tests in the runner's single process. I will look into this further.

PR #79, opened 2026-08-11
feat(materials): make Marmot and native EdelweissFE materials interchangeable

alexdummer:
Until now a Marmot material could only be reached *through* an element in EdelweissFE. Even the
single quadrature point wrapper, `MarmotMaterialHypoElasticWrapper`, was shaped like an element
too: it exchanged Ke/Pe vectors and owned its own state variable bookkeeping.

This adds `edelweissfe/materials/marmot/`, exposing Marmot materials as material points that
honor the very same interface as the native Python materials:

* `MarmotHypoElasticMaterial` implements `BaseHypoElasticMaterial`, i.e.
  `computeStress(stress, dStress_dStrain, dStrain, time, dTime)` plus the plane stress and
  uniaxial stress routines, all delegated to Marmot's own condensation.
* `MarmotGradientEnhancedHypoElasticMaterial` implements the new
  `BaseGradientEnhancedHypoElasticMaterial`, which mirrors Marmot's increment/response/tangents
  structs for a material with n nonlocal variables.

A Marmot material and a native EdelweissFE material are therefore interchangeable in any
consumer, and this PR demonstrates that interchangeability in both directions:

1. **A Marmot material through a native discretization.** The new `marmothypoelastic` provider
   lets an *ordinary* element (`provider=edelweiss`) drive a Marmot material via a plain
   `*material` block: `materialName` is the real Marmot material name, exactly as for every
   other provider, and `getMaterialClass` returns a `functools.partial` binding it, so it drops
   straight into the single-argument `materialClass(materialProperties)` call
   `AbqModelConstructor.createMaterialsFromInputFile` makes. `DisplacementElementMarmotHypoElastic`
   demonstrates it with a `CPS4` element on a small mesh; run against the identical mesh with the
   native `LinearElastic` material instead, the two agree to ~15 significant digits.
2. **A native material through the point-wise (QP) element.** `BaseMaterialDriver.setMaterial`
   lets a driver evaluate an already constructed material directly, not only a Marmot one it
   constructs itself from a name and properties (`createMaterial`). `QPEdelweissLinearElastic`
   and `QPEdelweissVonMises` demonstrate it: the same `MarmotSingleQpElement` mechanism the Marmot
   QP tests use, driving native `provider=edelweiss` materials instead.

The two point-wise extensions are registered under the provider `marmotmaterialpoint`; the
existing provider `marmotmaterial` keeps returning `None`, since that path means the Marmot
element owns the material. Each family's non-type template parameter (the number of nonlocal
variables, which Cython cannot express) is instantiated by a small C++ shim next to the sources,
which also keeps its own copy of the material properties, since Marmot itself only stores the
pointer.

**Element side**: the single quadrature point element now drives the point-wise materials
through a material driver per base class (`materialdrivers.py`), so its state variable layout
and its residual and tangent assembly live in one readable place. The gradient enhanced variant
is re-enabled: its residual is K - KLocal with the consistent tangent, whereas the old wrapper
had the sign of dKLocal/dStrain inverted and dropped dKLocal/dK entirely. The converged solution
is unchanged, i.e. the test passes against its original, untouched U.ref to within 2e-13.

**Performance**: the gradient-enhanced tangents now zero via a single contiguous-block fill
instead of six separate numpy operations (2.4 us down to 0.36 us per call, measured against
37 us spent per material point in a two dimensional gradient plasticity stencil in a sibling PR).

**Also bundled**:
* `domain=1d` models can now be built at all: `getFieldSize` only handled a rotation vector for
  `domainSize` 2 and 3, and `FEModel._createNodeFieldsFromNodes` asks for the field size of
  *every* declared phenomenon up front, so the one missing combination made any 1d model
  unbuildable. A one dimensional domain leaves a single rotation axis, exactly as a two
  dimensional one does, so it now returns 1.
* `setup.py` warns when the Marmot or Eigen headers are not found, since extension build
  failures are tolerated on purpose (`optional_build_ext`), which otherwise means a wrong
  include directory would only surface much later as a silently missing module.
* The point-wise Marmot material `.pxd` declarations and C++ shim headers are now shipped as
  package data (`MANIFEST.in` / `setup.py`), since a downstream package (e.g. EdelweissFD)
  compiling its own extensions against an installed, non-editable EdelweissFE needs to
  `cimport`/include them.

**Dead code removed**: `marmot.pxd`, `marmot.pyx` and
`marmotmaterialgradientenhancedhypoelasticwrapper.pyx` declared `Marmot/Marmot.h`,
`Marmot/MarmotMaterial.h`, `Marmot/MarmotMaterialGradientEnhancedHypoElastic.h` and
`MarmotMaterialFactory::getMaterialCodeFromName`, none of which exist in Marmot any more. Their
extension was already commented out in `setup.py` and their test already disabled. The compiled
finite-difference cell kernels (`displacementkernel.pyx`, `gradientenhanceddisplacementkernel.pyx`)
were also removed: nothing in EdelweissFE has any finite-difference solver or stencil machinery
to use them, they existed purely so EdelweissFD's stencils could delegate to them -- they now
live there instead ([alexdummer/EdelweissFD#1](https://github.com/alexdummer/EdelweissFD/pull/1)).

**Split into a stacked PR**: the gradient plasticity Marmot material family
(`MarmotMaterialGradientPlasticityHypoElastic`, its driver, base classes and C++ shim) is in
[#8](https://github.com/alexdummer/EdelweissFE/pull/8) (stacked on this branch), since no
available Marmot build here registers a concrete gradient plasticity material to test it
against, unlike the hypoelastic and gradient-enhanced families this PR demonstrates end to end.

## Testing
* `QPMarmotMaterialHypoElastic` and `QPMarmotMaterialHypoElasticVonMises` pass against their
  original, untouched `U.ref`.
* `QPMarmotMaterialGradientEnhancedAT2PhaseField` (new): drives the gradient-enhanced interface
  with a concrete Marmot material (AT2 phase field) end to end. The nonlocal damage variable
  stays exactly zero below a threshold then grows through a smooth sigmoid to ~1 as load
  increases, and the lateral/axial strain ratio matches the input Poisson's ratio at every
  increment -- the expected physical signature of isotropic degradation under uniaxial stress.
* `QPEdelweissLinearElastic` and `QPEdelweissVonMises` (new): the same single-quadrature-point
  setup, but driving native `provider=edelweiss` materials instead of Marmot ones, through the
  same driver. Both show physically correct behavior (lateral/axial strain ratio = Poisson's
  ratio for the elastic case; ~0.5 for VonMises at 20% strain, consistent with near-incompressible
  plastic flow).
* `DisplacementElementMarmotHypoElastic` (new): a small (2x2 `CPS4`) tension test with an
  ordinary `provider=edelweiss` element driven by a `provider=marmothypoelastic` material,
  cross-checked to ~15 significant digits against the same mesh with the native material.
* Merged `next_v26.11` in to resolve a `setup.py` conflict with its `EDELWEISSFE_ARCH_FLAGS`
  mechanism; re-ran the full `testfiles/edelweiss-only/` suite afterwards -- green except one
  pre-existing failure (`NodeToDiscreteRigidBodyContact`, a pyvista API mismatch unrelated to and
  untouched by this PR) -- and `pre-commit run --all-files`.
* Could not build or test the gradient-enhanced/gradient-plasticity extensions with every Marmot
  build available in this repo's own CI (no available build there registers a concrete
  gradient-plasticity material), but verified independently against a separate, complete Marmot
  install that does register one for the gradient-enhanced family (AT2 phase field).

https://claude.ai/code/session_01FRo5ZR4FKNdgyiSWAGQiBY


edelweissfe/materials/marmot/marmothypoelastic.pyx
```diff
@@ -0,0 +1,371 @@
...
+    cdef double[::1] _stateVars
+
+    cdef readonly str materialName
+
+    def __cinit__(self, materialName: str, materialProperties: np.ndarray, int materialNumber = 1):
+
+        self._material = NULL
+
+        self._materialProperties = np.ascontiguousarray(materialProperties, dtype=float)
+
```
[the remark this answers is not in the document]
alexdummer: Good catch, fixed in 9420ab9c: np.ascontiguousarray() alone returns the caller's own array object when it is already contiguous float64, so a later in-place mutation of it would have silently changed this material's behavior. Now forces an owned copy via np.array(..., copy=True).

edelweissfe/elements/marmotsingleqpelement/element.py
```diff
@@ -210,17 +212,17 @@ def computeKernelsExplicit(
     ):
         self._initializeStateVarsTemp()
 
-        self._marmotMaterialWrapper.computeKernelsExplicit(Pe, U, dU, time, dTime)
+        self._driver.computeKernels(np.zeros((self._nDof, self._nDof)), Pe, U, dU, time, dTime)
```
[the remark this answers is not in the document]
alexdummer: Fixed in 9420ab9c: added a preallocated self._KeScratch buffer reused across calls instead of allocating fresh zeros every time.

edelweissfe/elements/marmotsingleqpelement/materialdrivers.py
```diff
@@ -0,0 +1,481 @@
...
+
+        self._material.computeStress(self._stress, tangent, dStrain, time, dTime)
+
+        self._strain += dStrain
+
+        # stored column wise, so that a reshape with order='F' recovers the matrix
+        self._dStress_dStrain[:] = tangent.flatten(order="F")
+
+        Pe[:] = self._stress
+        Ke[:, :] = tangent
```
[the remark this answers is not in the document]
alexdummer: Fixed in 9420ab9c: Ke is now written to directly (it's already the caller's (6, 6) output buffer), removing the per-call tangent = np.zeros((6, 6)) allocation.

edelweissfe/config/materiallibrary.py
```diff
@@ -49,9 +49,40 @@ def getMaterialClass(materialName: str, provider: str = None) -> type:
...
+            from edelweissfe.materials.marmot.marmotgradientplasticityhypoelastic import (
+                MarmotGradientPlasticityHypoElasticMaterial,
+            )
+
+            return MarmotGradientPlasticityHypoElasticMaterial
+
+        raise Exception(
+            "Unknown Marmot material point base class '{:}'; expected 'hypoelastic', "
+            "'gradientEnhancedHypoElastic' or 'gradientPlasticityHypoElastic'".format(materialName)
+        )
```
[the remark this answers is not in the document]
alexdummer: Fixed in 9420ab9c: the message now uses the exact lowercase tokens the strCaseCmp checks compare against.

edelweissfe/materials/base/basegradientenhancedhypoelasticmaterial.py
```diff
@@ -0,0 +1,369 @@
...
+
+        Returns
+        -------
+        float
+            The density of the material."""
+
+    @abstractmethod
+    def getResult(self, result: str) -> np.ndarray:
+        """Get the result, as a persistent view which is continiously
+        updated by the material.
```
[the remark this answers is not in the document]
alexdummer: Fixed in 9420ab9c.

edelweissfe/materials/base/basegradientplasticityhypoelasticmaterial.py
```diff
@@ -0,0 +1,367 @@
...
+
+        Returns
+        -------
+        float
+            The density of the material."""
+
+    @abstractmethod
+    def getResult(self, result: str) -> np.ndarray:
+        """Get the result, as a persistent view which is continiously
+        updated by the material.
```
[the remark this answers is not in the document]
alexdummer: Fixed in 9420ab9c.

edelweissfe/materials/marmot/marmothypoelastic.pyx
```diff
@@ -0,0 +1,371 @@
...
+        if self._material == NULL:
+            raise ValueError(
+                "Marmot does not provide a hypoelastic material '{:}'".format(materialName)
+            )
+
+        self.materialName = materialName
+
+    def __dealloc__(self):
+        del self._material
```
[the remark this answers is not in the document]
alexdummer: This isn't actually unsafe: _material is set to NULL as the very first statement in __cinit__, before the Marmot factory call that could raise (and if that factory call itself raises, the assignment to self._material never completes, so it stays NULL). C++ delete on a null pointer is a well-defined no-op (never invokes the destructor, never crashes), so __dealloc__ can't segfault here. Leaving as-is.

edelweissfe/materials/marmot/marmotgradientenhancedhypoelastic.pyx
```diff
@@ -0,0 +1,333 @@
...
+            materialName_,
+            &self._materialProperties[0],
+            self._materialProperties.shape[0],
+            materialNumber,
+        )
+
+        self.materialName = materialName
+
+    def __dealloc__(self):
+        del self._material
```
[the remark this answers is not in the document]
alexdummer: Same as the note on the hypoelastic material's __dealloc__: _material is NULL-initialized before the constructor call that could raise, and if the 'new GradientEnhancedHypoElasticShim1(...)' call itself throws, the assignment never completes, so _material stays NULL. C++ delete on a null pointer is a well-defined no-op. Leaving as-is.

edelweissfe/materials/marmot/marmotgradientplasticityhypoelastic.pyx
```diff
@@ -0,0 +1,306 @@
...
+            materialName_,
+            &self._materialProperties[0],
+            self._materialProperties.shape[0],
+            materialNumber,
+        )
+
+        self.materialName = materialName
+
+    def __dealloc__(self):
+        del self._material
```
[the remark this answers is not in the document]
alexdummer: Same reasoning as the other two __dealloc__ comments: _material is NULL-initialized before the constructor call that could raise, so it stays NULL if construction fails, and C++ delete on a null pointer is a well-defined no-op. Leaving as-is.

alexdummer:
Re the Codecov patch-coverage report: both project and patch checks are configured as `informational: true` in codecov.yml (no baseline exists yet), and both are green (codecov/patch: SUCCESS) -- this isn't a merge gate.

The bulk of the remaining gap is the gradient-enhanced and gradient-plasticity code paths (materialdrivers.py's non-hypoelastic drivers, both base*hypoelasticmaterial.py files). These can't be exercised here: no available Marmot build, including CI's own, registers a concrete gradient-plasticity/gradient-enhanced material -- QPMarmotMaterialGradientEnhancedHypoElastic shows SKIPPED (module not found) in this PR's own CI run, same as noted in the branch's commit history. The hypoelastic point-wise path (materialdrivers.py's hypoelastic driver, marmothypoelastic.pyx, and materiallibrary.py's new dispatch) is exercised and passing via QPMarmotMaterialHypoElastic / QPMarmotMaterialHypoElasticVonMises.

alexdummer:
Added `testfiles/marmot/QPMarmotMaterialGradientEnhancedAT2PhaseField`, a QP regression test that drives the gradient-enhanced point-wise interface with a concrete Marmot material (AT2PhaseField) end to end, rather than only GCDP as `QPMarmotMaterialGradientEnhancedHypoElastic` does. Verified locally against a separate, complete Marmot install that registers AT2PhaseField: the nonlocal damage variable stays exactly zero below a threshold then grows through a smooth sigmoid to ~1 as load increases, and the lateral/axial strain ratio matches the input Poisson's ratio at every increment (expected under uniaxial stress with isotropic degradation) -- good evidence the point-wise interface is correct, not just non-crashing.

Like the existing `QPMarmotMaterialGradientEnhancedHypoElastic`, this will show as SKIPPED in this repo's current CI (neither this repo's CI Marmot checkout nor any other build available to me has the gradient-enhanced headers) -- not a regression, same known limitation already called out in the branch's commit history.

alexdummer:
Split the gradient plasticity material family out into a stacked PR: alexdummer#8 (`alex/marmot-gradient-plasticity` -> `alex/pointwise-marmot-materials`). This branch now only carries the hypoelastic and gradient-enhanced point-wise interfaces, both demonstrated end to end with concrete Marmot materials (LinearElastic/VonMises and AT2PhaseField respectively). Gradient plasticity had no available Marmot build to test against here, so it shouldn't hold up this branch; alexdummer#8 will get rebased onto `next_v26.11` once this merges.

edelweissfe/config/materiallibrary.py
```diff
@@ -64,9 +64,33 @@ def getMaterialClass(materialName: str, provider: str = None) -> type:
...
+    if strCaseCmp(provider, "marmotmaterialpoint"):
+        # A Marmot material evaluated point-wise, without an element in between.
+        # Which class applies follows from the base class the material is registered for
+        # in Marmot, not from its name, so the caller states the base class.
+        if strCaseCmp(materialName, "hypoelastic"):
+            from edelweissfe.materials.marmot.marmothypoelastic import (
+                MarmotHypoElasticMaterial,
+            )
+
+            return MarmotHypoElasticMaterial
```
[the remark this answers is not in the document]
alexdummer: Fair catch, fixed in ffde8e09: documented in the getMaterialClass docstring that marmotmaterialpoint is deliberately not usable via *material -- its materialName is a base-class token and the returned class needs (materialName, materialProperties), not the single materialProperties AbqModelConstructor passes. It exists for external code (e.g. a downstream finite-difference stencil) to look up the class and construct it directly, not for *material declarations.

edelweissfe/elements/marmotsingleqpelement/materialdrivers.py
```diff
@@ -0,0 +1,364 @@
...
+        if result == "algorithmicTangent":
+            return np.array(self._algorithmicTangent, copy=not getPersistentView)
+
+        return np.array(self._material.getResult(result), copy=not getPersistentView)
+
+
+#: The available material drivers, keyed by the name of the Marmot material base class.
+materialDrivers = {
+    "MarmotMaterialHypoElastic": MarmotMaterialHypoElasticDriver,
+    "MarmotMaterialGradientEnhancedHypoElastic": MarmotMaterialGradientEnhancedHypoElasticDriver,
```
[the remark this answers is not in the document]
alexdummer: This referred to the PR description before it was rewritten in ffde8e09's predecessor -- the gradient plasticity family and its driver are not part of this PR, they were split into a stacked PR (alexdummer#8) since no available Marmot build here registers a concrete gradient plasticity material to test against, unlike the two families this PR covers. The current description reflects that.

setup.py
```diff
@@ -91,35 +112,30 @@
...
-#         runtime_library_dirs=[join(marmot_dir, "lib")],
-#         language="c++",
-#     )
-# ]
+print("Gather the extensions for the point-wise Marmot material interfaces")
+marmot_material_dir = join("edelweissfe", "materials", "marmot")
+for marmot_material_source in [
+    "marmothypoelastic.pyx",
+    "marmotgradientenhancedhypoelastic.pyx",
+]:
```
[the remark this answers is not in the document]
alexdummer: Same as the other one -- this referred to the PR description before it was rewritten. The compiled finite-difference cell kernels were removed from this branch: nothing in EdelweissFE itself has any finite-difference solver/stencil machinery to use them, they existed purely so EdelweissFD's stencils could delegate to them, and now live there instead (alexdummer/EdelweissFD#1). The current description reflects that.

edelweissfe/materials/marmot/_marmotmaterials.pxd
```diff
@@ -43,44 +51,95 @@ cdef extern from "Marmot/MarmotTypedefs.h" nogil:
...
+    cdef double[::1] resultView = <double[:res.stateSize]> (res.stateLocation)
+
+    return np.asarray(resultView)
+
+
 cdef extern from "Marmot/MarmotMaterialHypoElastic.h":
     cdef cppclass MarmotMaterialHypoElastic nogil:
 
-        StateView getStateView(const string& stateName, double* stateVars) except +ValueError
+        StateView getStateView(const string& stateName, double* stateVars) except +
```
matthiasneuner: These exception declarations changed from `except +ValueError` to bare `except +` (same for the analogous ones in `_gradientenhanced.pxd`), but the `.pyx` call sites only catch `except (ValueError, RuntimeError)` before converting to a `CutbackRequest`.

Bare `except +` maps `std::out_of_range` → `IndexError`, `std::overflow_error` → `OverflowError`, `std::bad_alloc` → `MemoryError` — none of which match the narrow catch clause. `MarmotStateVarVectorManager` uses `std::map::at()`, which throws `std::out_of_range` on a state-var layout mismatch; that now escapes uncaught instead of triggering the intended graceful step cutback, crashing the increment.

edelweissfe/elements/marmotsingleqpelement/element.py
```diff
@@ -130,40 +137,48 @@ def setNodes(self, nodes: list):
...
         self,
     ):
         """
-        Not used by this wrapper"""
+        Not used by this driver"""
 
-    def setMaterial(self, materialName: str, materialProperties: np.ndarray):
-        """Assign a material and material properties to the underlying Wrapper.
-        Furthermore, create two sets of state vars:
+    def setMaterial(self, materialNameOrInstance, materialProperties: np.ndarray = None):
```
matthiasneuner, line 150: `setMaterial(materialNameOrInstance, materialProperties=None)` dispatches purely on `materialProperties is None`; previously both args were required, so a missing one raised `TypeError` immediately.

A caller invoking `element.setMaterial(materialName)` with only a positional string (an easy mistake now that the second arg is optional) no longer fails immediately — the bare string is stored as `self._material`, and the failure only surfaces later as a cryptic `AttributeError` like `'str' object has no attribute 'computeStress'`.

edelweissfe/elements/marmotsingleqpelement/element.py
```diff
@@ -210,17 +221,17 @@ def computeKernelsExplicit(
     ):
         self._initializeStateVarsTemp()
 
-        self._marmotMaterialWrapper.computeKernelsExplicit(Pe, U, dU, time, dTime)
+        self._driver.computeKernels(self._KeScratch, Pe, U, dU, time, dTime)
```
matthiasneuner, line 254: `computeKernelsExplicit` now calls `self._driver.computeKernels(self._KeScratch, Pe, U, dU, time, dTime)`, always computing the full `(nDof, nDof)` material tangent and immediately discarding it, unlike the native `MarmotElementWrapper`'s genuinely tangent-free explicit path.

Every explicit-dynamics timestep for every element of this type (called from `solvers/nonlinearexplicitdynamic.py` and `solvers/base/parallelelementcomputation.py`) pays for a full tangent evaluation it throws away — roughly double the per-call material cost for no benefit. (Note: the old code called `computeKernelsExplicit`, which the old wrapper never defined, so this path was previously broken/dead — this PR is what makes it functional, at this cost.)

matthiasneuner (commented):
Automated review pass (correctness, reuse/simplification/efficiency, docs). Two additional findings couldn't be attached as inline comments because they're on lines outside this PR's diff:

- **`edelweissfe/generators/abqmodelconstructor.py:311`** — `createAdvancedMaterialsFromInputFile()`'s guard against unsupported providers checks `materialClass is None`, which is true for `provider=marmotmaterial` but not for this PR's new `provider=marmothypoelastic` (`getMaterialClass` returns a `functools.partial`, never `None`). So `*advancedmaterial, provider=marmothypoelastic` silently skips the intended `raise Exception("Advanced materials are currently not possible with this material provider!")` and instead calls `partial(MarmotHypoElasticMaterial, NAME)(propertiesDict)` with a dict instead of an ndarray, raising a confusing low-level numpy/Cython error instead of the clear intended guard message.
- **Documentation** — this PR adds a new material-provider abstraction (`BaseGradientEnhancedHypoElasticMaterial`, the `marmothypoelastic`/`marmotmaterialpoint` providers, and the interchangeable-driver mechanism) but touches no files under `doc/`. `doc/source/documentation/materials.rst` already documents `BaseHypoElasticMaterial`/`BaseHyperElasticMaterial` via `autoclass` but isn't updated for the new base class/provider mechanism.

edelweissfe/elements/marmotsingleqpelement/materialdrivers.py
```diff
@@ -0,0 +1,383 @@
...
+    @property
+    def nDof(self) -> int:
+        return self._nDof
+
+    def createMaterial(self, materialName: str, materialProperties: np.ndarray):
+        from edelweissfe.materials.marmot.marmothypoelastic import (
+            MarmotHypoElasticMaterial,
+        )
+
+        self.setMaterial(MarmotHypoElasticMaterial(materialName, materialProperties))
```
matthiasneuner, line 199: Unlike `MarmotMaterialGradientEnhancedHypoElasticDriver.setMaterial` (which validates `nNonlocalVariables == 1`), `MarmotMaterialHypoElasticDriver` doesn't override `BaseMaterialDriver.setMaterial`, so nothing here validates that an already-constructed material actually implements the hypoelastic interface.

Handing a `MarmotGradientEnhancedHypoElasticMaterial` (or any `BaseGradientEnhancedHypoElasticMaterial`) to a `type=MarmotMaterialHypoElastic` element succeeds silently here, then fails deep in `computeKernels()` with a confusing `AttributeError` instead of a clear validation message.

edelweissfe/sets/elementset.py
```diff
@@ -35,7 +35,7 @@
 else:
     MarmotElementWrapper = None
 
-if checkSuccessfulExtension("edelweissfe.elements.marmotsingleqpelement.marmotmaterialhypoelasticwrapper"):
+if checkSuccessfulExtension("edelweissfe.materials.marmot.marmothypoelastic"):
```
matthiasneuner: `MarmotMaterialWrappingElement` is now gated only on `edelweissfe.materials.marmot.marmothypoelastic` having built successfully, but it also exposes `MarmotMaterialGradientEnhancedHypoElasticDriver`, backed by the separately-compiled `marmotgradientenhancedhypoelastic` extension.

If `marmothypoelastic` builds but `marmotgradientenhancedhypoelastic` fails (e.g. an older/partial Marmot install missing the general gradient-enhanced factory), the element is still exposed; `type=MarmotMaterialGradientEnhancedHypoElastic` then fails with a raw, unguarded `ImportError` from `materialdrivers.py`'s lazy import instead of the codebase's usual clean "extension not built" handling.

edelweissfe/materials/marmot/marmothypoelastic.pyx
```diff
@@ -0,0 +1,385 @@
...
+    # so this contiguous copy must outlive the material itself.
+    cdef double[::1] _materialProperties
+
+    cdef double[::1] _stateVars
+
+    cdef readonly str materialName
+
+    def __cinit__(self, materialName: str, materialProperties: np.ndarray, int materialNumber = 1):
+
+        self._material = NULL
```
matthiasneuner, line 97: `__cinit__`'s uppercase-retry fallback (on `NULL`) is unreachable: `MarmotMaterialHypoElasticFactory::createMaterial` always throws `std::invalid_argument` on an unregistered name rather than returning `nullptr`, so the outer `except Exception` fires first.

Downstream code constructing `MarmotHypoElasticMaterial` directly with a non-uppercase but valid name (e.g. `'linearelastic'`) gets a spurious `NotImplementedError` even though `'LINEARELASTIC'` is registered, since the intended case-insensitive retry never executes.

edelweissfe/config/phenomena.py
```diff
@@ -221,7 +221,8 @@ def getFieldSize(field, domainSize):
     if fType == "vector":
         return domainSize
     if fType == "rotation vector":
-        if domainSize == 2:
+        # a one and a two dimensional domain both leave a single rotation axis
+        if domainSize in (1, 2):
```
matthiasneuner, line 232: `getFieldSize()` for `fType == "rotation vector"` now returns 1 for `domainSize == 1` as well as 2; previously `domainSize == 1` fell through to the trailing `raise NotImplementedError`.

A `*job, domain=1d` model with a rotation-vector field (e.g. from a misconfigured/copy-pasted input) previously failed fast with a clear error; it now silently creates a 1-component rotation `NodeField` with no established physical justification for a 1D domain, masking what was previously a caught misconfiguration.

edelweissfe/materials/base/basegradientenhancedhypoelasticmaterial.py
```diff
@@ -0,0 +1,369 @@
...
+
+    def zero(self):
+        """Reset all entries to zero.
+
+        A single fill of the block the entries are views into, see :meth:`createZero`. Falls back
+        to entry by entry for an instance that was built by hand rather than by ``createZero``, so
+        that the class stays usable either way.
+        """
+
+        block = getattr(self, "_block", None)
```
matthiasneuner: `GradientEnhancedTangents.zero()` uses `getattr(self, "_block", None)` and, in its fallback loop, `getattr(self, name)` to read dataclass fields dynamically by string name.

Matthias' made up EdelweissFE conventions say to avoid `hasattr`/`getattr`/`setattr` constructs and instead use inheritance/polymorphism, with attributes and methods defined in the respective base classes. `_block` is set ad hoc in `createZero()` rather than as a declared dataclass field, which is what forces the existence check here and the by-name loop below.

edelweissfe/materials/marmot/_marmotmaterials.pxd
alexdummer: Confirmed and fixed in ea1985c7: restored except +ValueError on the declarations in both _marmotmaterials.pxd and _gradientenhanced.pxd, matching the pre-refactor marmot.pxd convention this replaced. Good catch -- a std::out_of_range from MarmotStateVarVectorManager::at() really would have escaped as an uncaught IndexError instead of the intended CutbackRequest.

edelweissfe/elements/marmotsingleqpelement/element.py
alexdummer: Fixed in ea1985c7: setMaterial now dispatches on isinstance(materialNameOrInstance, str) rather than only on whether materialProperties is None, and raises TypeError immediately (with a message naming the mistake) in both misuse directions -- a bare name with no properties, or an instance with properties supplied anyway.

edelweissfe/elements/marmotsingleqpelement/element.py
alexdummer: Checked Marmot's exposed interface (_marmotmaterials.pxd): none of computeStress/computePlaneStress/computeUniaxialStress offer a tangent-free variant, so there's no cheaper call available on the EdelweissFE side without a Marmot-side API addition, which is out of scope here. Documented the trade-off explicitly on computeKernelsExplicit's docstring in ea1985c7 rather than pretending to fix it, per your own note that this path was previously dead code anyway.

edelweissfe/elements/marmotsingleqpelement/materialdrivers.py
alexdummer: Fixed in ea1985c7: both MarmotMaterialHypoElasticDriver and MarmotMaterialGradientEnhancedHypoElasticDriver now validate in setMaterial() that the material is an instance of the base class they expect (isinstance check against BaseHypoElasticMaterial / BaseGradientEnhancedHypoElasticMaterial), raising a clear TypeError instead of failing deep in computeKernels().

edelweissfe/sets/elementset.py
alexdummer: Fixed in ea1985c7: the gate now exposes MarmotMaterialWrappingElement if either extension built, since it was previously wrong in both directions (hidden entirely when only marmotgradientenhancedhypoelastic built; silently missing the check on that extension otherwise). Also documented on BaseMaterialDriver.createMaterial's docstring that a driver whose own extension isn't built raises ImportError from its lazy import -- matching the existing convention elsewhere (e.g. linsolve/klu/__init__.py) rather than trying to wrap/rename it.

edelweissfe/materials/marmot/marmothypoelastic.pyx
alexdummer: Confirmed and fixed in ea1985c7: split into two independent try/except blocks (the retry is now only attempted after the first one has already failed), matching the pattern _gradientenhancedshim.h already used correctly. Verified directly: MarmotHypoElasticMaterial('linearelastic', ...) now succeeds instead of raising a spurious NotImplementedError.

edelweissfe/config/phenomena.py
alexdummer: Fair concern, and worth being precise about: getFieldSize is called unconditionally for every *declared* phenomenon by _createNodeFieldsFromNodes, regardless of whether any element in the model actually uses it -- an unused field is filtered out afterwards there (its NodeField ends up with zero nodes), not here. So this needed to return *some* valid size for domainSize == 1 just to keep every 1D model buildable (the original motivation for this PR), not because a rotation vector is a physically meaningful field on a 1D domain -- no element in this codebase currently declares one. Reworded the comment in ea1985c7 to say that plainly rather than implying a physical justification that isn't really there. Reverting to NotImplementedError for domainSize == 1 specifically would re-break every 1D model again, including ones that never touch rotation vectors at all, so I kept the behavior and fixed the reasoning in the comment instead.

edelweissfe/materials/base/basegradientenhancedhypoelasticmaterial.py
alexdummer: Fixed in ea1985c7: _block is now a declared dataclass field (defaulting to None) rather than an attribute createZero() bolts on afterwards, so zero() can tell the two construction paths apart with a plain self._block is not None check. Dropped entryNames entirely (it had no other use) and iterate over the actual field references directly in the fallback loop instead of getattr(self, name).

Issue #82, opened 2026-08-14
Hexa8/Hexa20 local node ordering is not Abaqus-compatible -- blocks real Abaqus mesh import


matthiasneuner:
## Summary

Discovered while triaging a review finding on #57 about `surfaceElementGenerator`'s face-numbering
convention. The scope of that finding turned out to be bigger than just contact-surface face
numbers: this codebase's own `Hexa8` element uses a genuinely different **local node ordering**
than the real Abaqus `C3D8` convention -- not just a different face-labeling scheme layered on top
of the same ordering.

## Evidence

`edelweissfe/elements/displacementelement/_elementcomputationmatrices.py`'s `computeNOperator`
defines `Hexa8`'s shape functions. Reverse-engineering each shape function's own natural coordinate
`(xi, eta, zeta)` (the corner where it evaluates to 1) gives this codebase's actual local node
ordering, compared against the standard Abaqus `C3D8` convention:

| slot | this codebase's Hexa8 | real Abaqus C3D8 |
|---|---|---|
| 1 | (-1,-1,-1) | (-1,-1,-1) |
| 2 | (-1,-1,+1) | (+1,-1,-1) |
| 3 | (+1,-1,+1) | (+1,+1,-1) |
| 4 | (+1,-1,-1) | (-1,+1,-1) |
| 5 | (-1,+1,-1) | (-1,-1,+1) |
| 6 | (-1,+1,+1) | (+1,-1,+1) |
| 7 | (+1,+1,+1) | (+1,+1,+1) |
| 8 | (+1,+1,-1) | (-1,+1,+1) |

They only agree at slots 1 and 7. This codebase's own ordering groups nodes into two four-node
rings split by `eta` (matching `boxGen`'s own "1=Ymin, 2=Ymax, ..." face-numbering pattern, see
`edelweissfe/generators/boxgen.py` and the face-numbering comment in
`edelweissfe/generators/surfaceelementgenerator.py`); Abaqus's groups them into two rings split by
`zeta` instead.

## Impact

If a genuine external Abaqus `.inp` file's `*element, type=C3D8` connectivity (node labels listed
in real Abaqus order) were fed into this codebase as-is, the resulting element would be a
geometrically distorted/twisted hexahedron, not merely have mislabeled contact faces. This means:

- The existing, narrower gap (real Abaqus `S1..S6` face numbers on a `*surface` keyword being
  misinterpreted under this codebase's own face-numbering convention, see
[...]

PR #83, opened 2026-08-15
fix(elements): unify Hexa8/Hexa20 node ordering with Marmot's Abaqus-standard convention

matthiasneuner:
## Summary

The pure-Python `Hexa8`/`Hexa20` elements (`displacementelement` and `displacementtlelement`) used
a local node ordering that split the two corner rings by the second natural coordinate (eta),
differing from Marmot's genuine Abaqus C3D8/C3D20 convention (rings split by zeta). Both
conventions were internally self-consistent on their own, but this meant the framework had two
silently-different node orderings for the same nominal element type ("hexa8") depending on which
provider backed it -- confirmed directly against Marmot's C++ shape functions
(`MarmotFiniteElementCore/src/MarmotFiniteElement3D.cpp`).

This surfaced while investigating a `surfaceElementGenerator` face-numbering review finding on
#57 (contact-facet generation) -- the actual defect turned out to be one level deeper, in the
element formulation itself, not just the facet-generator's face-number lookup.

## What changed

- `computeNOperator` and the Jacobian/B-operator machinery for `Hexa8`/`Hexa20` (both
  `displacementelement` and `displacementtlelement`) now use Marmot's node ordering.
- The old Jacobian/B-operator code hardcoded a full closed-form polynomial per node order
  (duplicated across `_J3D8`, `_B3D8`, `_J3D20`, `_B3D20`, and again in the TL variant) -- this is
  exactly the kind of silent-drift risk that caused the mismatch in the first place. Replaced with
  a single, shared dN/dxi-based generic isoparametric construction, so the node ordering only needs
  to be correct in one place (`computeNOperator`/the dNdXi helper) going forward.

## What did *not* need to change (verified, not assumed)

- `boxGen`/`pipeGen`'s own node-generation loops -- confirmed numerically that Marmot's real shape
  functions, applied directly to boxgen's *unchanged* connectivity, give a valid, positive-Jacobian
  element (the two conventions are related by a consistent, orientation-preserving axis
  permutation). boxGen's own face-number registration (1=Ymin, 2=Ymax, 3=Xmin, 4=Zmax, 5=Xmax,
  6=Zmin) also already matches Marmot's own face IDs exactly.
- The one hand-written hex mesh in the test suite (`CantileverBeamHexa8/test.inp`+`mesh.inp`, 160
  explicit `C3D8` elements) -- checked the same way, needs no reordering.

## Verification

- Patch tests: constant strain field correctly reproduced by the new B-operator on a skewed
  (non-cubic) hexahedron, both `Hexa8` and `Hexa20`.
- Full `testfiles/edelweiss-only/` suite: 0 failures (results are physically identical to before,
  as expected for a pure internal relabeling on boxgen/pipegen-generated meshes).

## Related

- Closes/supersedes [#82](https://github.com/Edelweiss-Numerics/EdelweissFE/issues/82).
- #57 cherry-picks this commit and re-derives its own `_FACE_TABLES` on top of it.

### DIFF
--- a/doc/source/documentation/elements.rst
+++ b/doc/source/documentation/elements.rst
@@ -46,7 +46,7 @@ One of the following element types needs to be included in the definition (``typ
 
 .. note::
     ``C3D8``/``C3D20`` (and their ``TL`` counterparts) use the standard Abaqus node ordering
-    (corner ring 0-3 at :math:`\zeta=-1`, ring 4-7 at :math:`\zeta=+1`), matching the ``marmot``
+    (corner ring 1-4 at :math:`\zeta=-1`, ring 5-8 at :math:`\zeta=+1`), matching the ``marmot``
     provider's elements -- there is only one node-ordering convention across the framework.
 
 **additional Parameters**
--- a/doc/source/documentation/elements.rst
+++ b/doc/source/documentation/elements.rst
@@ -44,6 +44,11 @@ One of the following element types needs to be included in the definition (``typ
 - **C3D8**    - hexahedron 3D element with 8 nodes.
 - **C3D20**    - hexahedron 3D element with 20 nodes.
 
+.. note::
+    ``C3D8``/``C3D20`` (and their ``TL`` counterparts) use the standard Abaqus node ordering
+    (corner ring 0-3 at :math:`\zeta=-1`, ring 4-7 at :math:`\zeta=+1`), matching the ``marmot``
+    provider's elements -- there is only one node-ordering convention across the framework.
+
 **additional Parameters**
 
 The following optional Parameters are also included in the element type definition:
--- a/edelweissfe/elements/displacementelement/_elementcomputationmatrices.py
+++ b/edelweissfe/elements/displacementelement/_elementcomputationmatrices.py
@@ -652,7 +652,7 @@ def _B2D8(xi: np.ndarray, eta: np.ndarray, x: np.ndarray, nInt: int):
     return Bi
 
 
-def _B3DGeneric(dNdXi: np.ndarray, x: np.ndarray, J: np.ndarray, nInt: int, nNodes: int):
+def _B3DGeneric(dNdXi: np.ndarray, J: np.ndarray, nInt: int, nNodes: int):
     """Assemble the standard isoparametric strain-displacement (B) operator for a 3D solid
     element from its dN/dxi and Jacobian, given a node-major local DOF vector
     ``(ux1,uy1,uz1,ux2,uy2,uz2,...)``.
@@ -661,8 +661,6 @@ def _B3DGeneric(dNdXi: np.ndarray, x: np.ndarray, J: np.ndarray, nInt: int, nNod
     ----------
     dNdXi
         Shape ``(nInt, 3, nNodes)``, as returned by :func:`_dNdXi3D8`/:func:`_dNdXi3D20`.
-    x
-        Coordinates of the element points.
     J
         The Jacobian matrices, shape ``(nInt, 3, 3)``.
     nInt
@@ -678,7 +676,7 @@ def _B3DGeneric(dNdXi: np.ndarray, x: np.ndarray, J: np.ndarray, nInt: int, nNod
 
     Bi = np.zeros([nInt, 6, 3 * nNodes])
     for i in range(nInt):
-        dNdX = np.linalg.inv(J[i]) @ dNdXi[i]  # (3, nNodes): row b = dN/dX_b
+        dNdX = np.linalg.solve(J[i], dNdXi[i])  # (3, nNodes): row b = dN/dX_b
         for k in range(nNodes):
             dNkdX, dNkdY, dNkdZ = dNdX[:, k]
             Bi[i, :, 3 * k : 3 * k + 3] = np.array(
@@ -717,7 +715,7 @@ def _B3D8(xi: np.ndarray, eta: np.ndarray, z: np.ndarray, x: np.ndarray, nInt: i
 
     dNdXi = _dNdXi3D8(xi, eta, z, nInt)
     J = _J3D8(xi, eta, z, x, nInt)
-    return _B3DGeneric(dNdXi, x, J, nInt, 8)
+    return _B3DGeneric(dNdXi, J, nInt, 8)
 
 
 def _B3D20(xi: np.ndarray, eta: np.ndarray, z: np.ndarray, x: np.ndarray, nInt: int):
@@ -743,4 +741,4 @@ def _B3D20(xi: np.ndarray, eta: np.ndarray, z: np.ndarray, x: np.ndarray, nInt: 
 
     dNdXi = _dNdXi3D20(xi, eta, z, nInt)
     J = _J3D20(xi, eta, z, x, nInt)
-    return _B3DGeneric(dNdXi, x, J, nInt, 20)
+    return _B3DGeneric(dNdXi, J, nInt, 20)
--- a/edelweissfe/elements/displacementelement/_elementcomputationmatrices.py
+++ b/edelweissfe/elements/displacementelement/_elementcomputationmatrices.py
@@ -28,6 +28,13 @@
 
 import numpy as np
 
+from edelweissfe.elements._hexa3dnodeordering import (
+    hexa8DNdXi,
+    hexa8N,
+    hexa20DNdXi,
+    hexa20N,
+)
+
 
 def computeJacobian(xi: np.ndarray, eta: np.ndarray, z: np.ndarray, x: np.ndarray, nInt: int, nNodes: int, dim: int):
     """Get the Jacobi matrix for the element calculation.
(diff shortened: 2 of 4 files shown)


edelweissfe/elements/displacementelement/_elementcomputationmatrices.py
```diff
@@ -676,6 +652,48 @@ def _B2D8(xi: np.ndarray, eta: np.ndarray, x: np.ndarray, nInt: int):
...
+    -------
+    np.ndarray
+        The B operator, shape ``(nInt, 6, 3*nNodes)``, Voigt order
+        ``(exx, eyy, ezz, gxy, gxz, gyz)``."""
+
+    Bi = np.zeros([nInt, 6, 3 * nNodes])
+    for i in range(nInt):
+        dNdX = np.linalg.inv(J[i]) @ dNdXi[i]  # (3, nNodes): row b = dN/dX_b
+        for k in range(nNodes):
+            dNkdX, dNkdY, dNkdZ = dNdX[:, k]
```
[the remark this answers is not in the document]
matthiasneuner: Fixed in `cc37efd9`: replaced `np.linalg.inv(J[i]) @ dNdXi[i]` with `np.linalg.solve(J[i], dNdXi[i])`, which computes the same result without explicitly forming the inverse. Full suite re-verified afterward (0 failures).

doc/source/documentation/elements.rst
```diff
@@ -44,6 +44,11 @@ One of the following element types needs to be included in the definition (``typ
 - **C3D8**    - hexahedron 3D element with 8 nodes.
 - **C3D20**    - hexahedron 3D element with 20 nodes.
 
+.. note::
+    ``C3D8``/``C3D20`` (and their ``TL`` counterparts) use the standard Abaqus node ordering
+    (corner ring 0-3 at :math:`\zeta=-1`, ring 4-7 at :math:`\zeta=+1`), matching the ``marmot``
+    provider's elements -- there is only one node-ordering convention across the framework.
```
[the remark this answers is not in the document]
matthiasneuner: Fixed in `cc37efd9`: the doc note now says "ring 1-4 at zeta=-1, ring 5-8 at zeta=+1" (1-based, matching Abaqus's own convention). Left the code comment in `_elementcomputationmatrices.py` 0-based, since that one describes actual array indices rather than user-facing Abaqus node numbers.

edelweissfe/elements/displacementelement/_elementcomputationmatrices.py
```diff
@@ -676,6 +652,48 @@ def _B2D8(xi: np.ndarray, eta: np.ndarray, x: np.ndarray, nInt: int):
...
+    element from its dN/dxi and Jacobian, given a node-major local DOF vector
+    ``(ux1,uy1,uz1,ux2,uy2,uz2,...)``.
+
+    Parameters
+    ----------
+    dNdXi
+        Shape ``(nInt, 3, nNodes)``, as returned by :func:`_dNdXi3D8`/:func:`_dNdXi3D20`.
+    x
+        Coordinates of the element points.
+    J
```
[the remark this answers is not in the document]
matthiasneuner: Fixed in `cc37efd9`: dropped the unused `x` parameter from `_B3DGeneric` entirely (dNdX is derived from `J` and `dNdXi` alone) and updated both call sites (`_B3D8`, `_B3D20`). Agreed removing it is cleaner than renaming to `_x` here, since it's genuinely never needed rather than intentionally-ignored-but-required-by-a-shared-signature.

edelweissfe/elements/displacementelement/_elementcomputationmatrices.py
```diff
@@ -161,6 +161,10 @@ def computeNOperator(xi: np.ndarray, eta: np.ndarray, z: np.ndarray, nInt: int,
                     )
                 )
     elif dim == 3:
+        # Hexa8/Hexa20 node ordering follows the standard Abaqus C3D8/C3D20 convention (corner
+        # ring 0-3 at zeta=-1, ring 4-7 at zeta=+1), matching Marmot's element formulation --
+        # see edelweissfe.generators.surfaceelementgenerator's module docstring for why this
```
alexdummer: Nit: this comment points to `edelweissfe.generators.surfaceelementgenerator`'s module docstring, but no such module exists in the repo (checked via grep/find). Either the module was never added, or the reference is stale — worth fixing so the rationale for the ordering convention stays traceable.

edelweissfe/elements/displacementtlelement/_elementcomputationmatrices.py
```diff
@@ -193,6 +193,10 @@ def computeNOperator(xi: np.ndarray, eta: np.ndarray, z: np.ndarray, nInt: int,
                     )
                 )
     elif dim == 3:
+        # Hexa8/Hexa20 node ordering follows the standard Abaqus C3D8/C3D20 convention (corner
+        # ring 0-3 at zeta=-1, ring 4-7 at zeta=+1), matching Marmot's element formulation and
+        # the small-strain displacementelement -- see edelweissfe.generators.surfaceelementgenerator's
```
alexdummer: Same dangling reference here — `edelweissfe.generators.surfaceelementgenerator` doesn't exist in the repo. This comment was copy-pasted from the displacementelement file, so the broken reference exists in two places now.

edelweissfe/elements/displacementtlelement/_elementcomputationmatrices.py
```diff
@@ -386,41 +390,42 @@ def _Ndiff3D(xi: np.ndarray, eta: np.ndarray, z: np.ndarray, nNodes: int):
     np.ndarray
         The derivative of the 3D shape functions w.r.t. the local coordinates."""
 
+    # Node ordering matches computeNOperator (Marmot's/standard Abaqus C3D8/C3D20 convention).
```
alexdummer, line 359: This PR's stated goal is eliminating 'two silently-different node orderings for the same element type', but the Hexa8/Hexa20 shape-function-derivative formulas are still hardcoded independently in two places: `_dNdXi3D8`/`_dNdXi3D20` in `displacementelement/_elementcomputationmatrices.py`, and `_Ndiff3D` here. If a future change (e.g. a reduced-integration variant, or another reordering) is applied to one but not the other, the small-strain and TL Hexa8/Hexa20 elements can silently diverge again — the same class of bug this PR fixed. Worth sharing one implementation across both files.

edelweissfe/elements/displacementelement/_elementcomputationmatrices.py
```diff
@@ -697,128 +713,13 @@ def _B3D8(xi: np.ndarray, eta: np.ndarray, z: np.ndarray, x: np.ndarray, nInt: i
...
-        [
-            [1, 0, 0, 0, 0, 0, 0, 0, 0],
-            [0, 0, 0, 0, 1, 0, 0, 0, 0],
-            [0, 0, 0, 0, 0, 0, 0, 0, 1],
-            [0, 1, 0, 1, 0, 0, 0, 0, 0],
-            [0, 0, 1, 0, 0, 0, 1, 0, 0],
-            [0, 0, 0, 0, 0, 1, 0, 1, 0],
-        ]
-    )
+    dNdXi = _dNdXi3D8(xi, eta, z, nInt)
```
alexdummer, line 594: Minor inefficiency: `_B3D8` computes `dNdXi` directly and then calls `_J3D8`, which recomputes the identical `dNdXi` internally via `_dNdXi3D8`. Same pattern likely applies to `_B3D20`/`_J3D20`. Harmless for correctness, but `_J3D8`/`_J3D20` could just accept the already-computed `dNdXi` instead of re-deriving it, avoiding the duplicate evaluation on every element call.

alexdummer:
Maybe implement a unified elementmatrices bor both small strain adn TL formulation?

edelweissfe/elements/displacementelement/_elementcomputationmatrices.py
matthiasneuner: Good catch, fixed in `1e4aa99f`: reworded to describe the reason directly (this is the local node ordering assumed by any face-numbering convention built on top of this element) rather than pointing to a specific module -- that module belongs to a separate PR (#57) not present on this branch.

edelweissfe/elements/displacementtlelement/_elementcomputationmatrices.py
matthiasneuner: Fixed the same way in `1e4aa99f`. Also addressed the underlying duplication this copy-paste symptom was pointing at -- see the reply below on your other comment.

edelweissfe/elements/displacementtlelement/_elementcomputationmatrices.py
matthiasneuner: Agreed, and fixed in `1e4aa99f`: extracted both the N and dNdXi formulas (Hexa8 and Hexa20) into a new shared module, edelweissfe.elements._hexa3dnodeordering, and made computeNOperator/_dNdXi3D8/_dNdXi3D20 (small-strain) and computeNOperator/_Ndiff3D (TL) thin wrappers around it. Verified numerically identical to the previous per-file formulas before swapping, for both N and dNdXi, both element types, both files. Full suite unaffected (0 failures).

edelweissfe/elements/displacementelement/_elementcomputationmatrices.py
matthiasneuner: Fixed in `1e4aa99f`: added a small `_JFromDNdXi(dNdXi, x, nInt)` helper. `_B3D8`/`_B3D20` now compute `dNdXi` once and pass it directly to `_JFromDNdXi`, instead of also calling `_J3D8`/`_J3D20` (which would recompute it internally). `_J3D8`/`_J3D20` themselves are thin wrappers around the same helper, kept standalone since `computeJacobian`'s dispatcher still needs them independently of `_B3D8`/`_B3D20`.

PR #91, opened 2026-08-25
feat(linsolve): unify all solvers behind a LinearSolver contract, add field-split block-AMG and offline benchmarking

matthiasneuner:
## Summary

Lands the linear-solver investigation campaign, consolidated onto current `next_v26.11`. Three
things, of which the **first touches every linear solver in the tree** — flagged up front because
that is not obvious from a squashed diff:

**1. A shared `LinearSolver` base contract, adopted by all 11 registered solvers.**
New `edelweissfe/linsolve/base.py` defines the interface every `linsolver` now presents:
`__call__(A, b) -> x`, plus `setJournal()`, `setModel()`, `setFieldStructure()` and the phase-split
`factorize()`/`solveFactorized()`, all with safe defaults. This replaced a growing pile of
per-capability setters and let the nonlinear solvers stop special-casing individual solvers.

Every pre-existing solver was migrated to it. The per-solver diffs are small and mechanical
(26–39 changed lines each) — each factory now returns a `LinearSolver` subclass instead of a bare
lambda, a bound method, or an option-constructed object:

| Solver | Change |
|---|---|
| `superlu`, `umfpack` | inline SciPy lambda → `LinearSolver` subclass |
| `pardiso` | wrapper giving the Cython extension type the base interface, and forwarding the phase-split API |
| `panuapardiso`, `klu`, `petsclu`, `mumps` | → `LinearSolver` subclass, construction-time import errors preserved |
| `amgcl` | → `LinearSolver` subclass |
| `gmres` | existing `Gmres` class gains `__call__` and inherits the base |

The `(A, b) -> x` calling convention is unchanged, so no `.inp` file changes meaning.

**2. New solvers and the AMGCL work:** an AMGCL-backed LGMRES outer solver with AMGCL-threaded MPC
condensation, and `blockamg` — a field-split block-AMG solver for large coupled multi-field systems
(vectorized in-place diagonal equilibration, threaded off-diagonal SpMV, opt-in gap-compensated
Eisenstat–Walker forcing tolerance).

**3. Offline benchmarking:** `matrixdump` dumps the equation systems handed to it and delegates the
actual solve, so solver variants are compared offline on one authentic, byte-identical matrix
sequence via `scripts/benchmark_linsolve.py` instead of by rerunning the simulation.

> **On splitting this up** (per @alexdummer's review): item 1 is not cleanly extractable, because
> items 2 and 3 are what motivated the contract and are the only reason several of its methods
> exist (`setModel` for `blockamg`'s field split and near-null-space; the phase-split API for the
> `lagged` benchmark). Splitting would produce a first PR of interface churn with no consumer, and
> a second that immediately rewrites it. Surfaced in the title instead, and itemised above so the
> squashed commit stays traceable.

Supersedes `perf/blockamg-efficiency` and `perf/blockamg-continuation-eta` as standalone PRs — all
of their content is included here.

## How this was built

The source branch (`perf/linsolve-consolidated`) forked before most of the recent stack (`#54`
through `#89`) and would produce dozens of ghost `add/add` conflicts if merged/rebased directly —
the same shape already seen with `#57`/`#61`/`feat/amr-recovery-marker`. Instead, this diffs only
the 114 commits genuinely unique to the linsolve work (everything past `feat/amr-recovery-marker`'s
original tip, already merged as `#89`) against current `next_v26.11`. Two real content conflicts
surfaced and were resolved by combining both sides:

- `edelweissfe/config/registry.py`: trunk's neo-Hookean material consolidation (`material` count
  11→7) and this branch's new linsolver registrations (`blockamg`/`matrixdump`) touched the same
  doc-count line; kept both, correcting the total to 11.
- `edelweissfe/linsolve/superlu/__init__.py`: trunk did comment-only cleanup; this branch
  refactored `createSolver()` to return a `SuperLUSolver(LinearSolver)` instance — took the
  branch's structural version.

Also caught (not a git conflict — a semantic one, since it's a `.inp` file referencing a material
by name): the new `CantileverBeamQuad8NeoHookeInexactNewton` regression test used the
pre-consolidation material name `neohookewaplastic`. Updated it to the unified `neohookeplastic`
class with its formulation-selector parameter; the test then passes against its own pre-existing
`U.ref` unchanged, confirming the consolidation preserved the Wa formulation's numerics exactly.

## Test plan

- [x] `pip install -v -e .` — builds cleanly except the pre-existing, unrelated `marmotelement`/
      `panuapardiso`/`klu` gaps (stale installed Marmot missing the named-element-properties API;
      documented, out of scope here)
- [x] Full `testfiles/edelweiss-only/` suite: 0 regressions (1 pre-existing failure, `MeshPlot`,
      needs a working `latex` toolchain this machine lacks — unrelated to this change)
- [x] The two new tests this branch adds (`CantileverBeamQuad4BlockAMG`,
      `CantileverBeamQuad8NeoHookeInexactNewton`) both pass
- [x] Registry smoke-check: all 11 `linsolver` names resolve via `edelweissfe.config.registry.lookup`
- [x] `black`/`isort`/`flake8` clean on all hand-resolved files
- [ ] `testfiles/marmot/` AMR/linsolve-relevant subset — blocked on this machine by the pre-existing
      stale-Marmot issue above; needs a machine with a current Marmot build to confirm

---

### Scope change during review

`inexactnewton` was **dropped** from this PR (see the discussion on
[this thread](https://github.com/Edelweiss-Numerics/EdelweissFE/pull/91#discussion_r3856164554)).
Three verified reasons: nothing used it outside its own regression test; it duplicated `blockamg`'s
Eisenstat–Walker forcing and residual-jump increment detection (the increment-detection expression
was byte-identical between the two files, which is why the zero-residual guard existed in one copy
and not the other); and it is a modified-Newton–Krylov *nonlinear* strategy wearing a linear
solver's interface, inferring increment boundaries from a jump in `||b||` because it has no access
to the Newton loop. The measured lagged-LU result stands as a finding and remains reachable offline
via `benchmark_linsolve.py`'s `lagged` subcommand — which is why PARDISO's phase-split
`factorize()`/`solveFactorized()` API is kept.

So this PR registers **two** new linsolvers (`blockamg`, `matrixdump`), for a registry total of 11.

Also added during review: a dedicated `blockamgtheory` documentation page (theory, all 29
parameters, solve-flow diagram) nested under Linear solvers, and fixes for six review findings on
`blockamg`/`pardiso`/`NEST` — see the commit log.


edelweissfe/drivers/inputfiledrivensimulation.py
```diff
@@ -173,18 +173,29 @@ def finiteElementSimulation(
...
-                [
-                    ("Step computation time", "{:10.4f}s".format(stepTime)),
-                ],
-                identification,
-                level=0,
-            )
+            try:
+                step.solve()
+            finally:
+                # Accumulate and report elapsed time for this step regardless of whether it
```
matthiasneuner: Condense overly verbose comments

edelweissfe/linsolve/inexactnewton/inexactnewton.py
```diff
@@ -0,0 +1,365 @@
...
+    def _forcingTolerance(self, residualNorm: float) -> float:
+        """The Eisenstat--Walker "choice 2" forcing tolerance for a reuse solve, clamped and safeguarded.
+
+        ``eta_k = gamma (||b_k|| / ||b_{k-1}||) ** alpha``, with the standard safeguard that prevents
+        the tolerance from tightening faster than the nonlinear residual actually warrants, then
+        clamped to ``[etaMin, etaMax]``.
+        """
+
+        ratio = residualNorm / self._lastResidualNorm
+        eta = self._ewGamma * ratio**self._ewAlpha
```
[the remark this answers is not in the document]
matthiasneuner: Fixed — added the same zero-residual guard blockamg._forcingTolerance already carries (falls back to etaMax when the previous residual was exactly zero).

edelweissfe/numerics/mpctransformation.py
```diff
@@ -199,25 +249,103 @@ def checkDirichletConflicts(self, dirichletDofIndices: np.ndarray):
...
+        ``product()``/``sum()`` instead of SciPy's single-threaded CSR sparse routines -- see
+        ``useAmgclSpgemm``'s own constructor-argument docstring for the measured speedup and the
+        ``eliminate_zeros()`` rationale.
+        """
+        from edelweissfe.linsolve.amgcl.amgcl import PyAMGCLSpGEMM
+
+        helper = PyAMGCLSpGEMM()
+        KT = helper.product(K, self._T)
+        KtNoC = helper.product(self._T.T.tocsr(), KT)
+        Kt = helper.sum(1.0, KtNoC, 1.0, self._C)
```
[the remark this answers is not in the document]
matthiasneuner: Fixed — self._T is assigned once in __init__ and never reassigned, so both the PyAMGCLSpGEMM helper and self._T.T.tocsr() are now cached on self and built lazily on first use.

edelweissfe/drivers/inputfiledrivensimulation.py
matthiasneuner: Fixed — condensed the 9-line comment to 3 lines, kept just the essential why (finally so a step that raises still counts its elapsed time).

examples/AnchorPryOut/zsymm_free_block.txt
matthiasneuner, line 1: Why do we need this file? Shouldn't it be generated on the file?

edelweissfe/linsolve/inexactnewton/inexactnewton.py
matthiasneuner, line 1: Why is this a separate file?

edelweissfe/solvers/nonlinearimplicitstatic.py
```diff
@@ -247,8 +289,91 @@ def solveStep(
...
+                    # re-pushed here on every (re)build so it tracks the mesh across AMR.
+                    self.linSolver.setModel(model, self.theDofManager)
+
+                    # Optional one-time dump of nodal coordinates aligned with the DOF vector, for
+                    # offline preconditioner experiments that need geometry (e.g. replaying a captured
+                    # system through a solver driven directly, outside a live model). The condensed
+                    # system keeps the DofManager ordering (MPC transform is size-preserving), so a
+                    # field's node coordinates in `field.nodes` order line up 1:1 with its DOF slice.
+                    # Gated by an environment variable so it never runs in production; overwrites on
+                    # every (re)build so the file reflects the current mesh after any AMR.
```
matthiasneuner: Why is this a responsibility of the nonlinear solver (instead of the linear solver?)

edelweissfe/solvers/nonlinearimplicitstatic.py
```diff
@@ -247,8 +289,91 @@ def solveStep(
...
+                    # Optional dump of the corner/midside topology of every vector field (the
+                    # p-multigrid enabler): the classification a P1 restriction operator needs
+                    # (identity on corners, 1/2-1/2 on each exclusive midside from its two
+                    # edge-endpoint corners), in the same field-node order the coords dump above
+                    # already uses. Scalar fields (e.g. nonlocal damage) have no P1-vs-quadratic story
+                    # of their own and are skipped. Gated by an environment variable, same reasoning as
+                    # the coordinate dump above -- a solver wanting a P1 map for its own use (e.g.
+                    # blockamg's p1FieldNames option) now computes it lazily itself, from the model
+                    # reference setModel already gave it above, rather than needing it pushed here.
+                    p1MapDumpDir = os.environ.get("EDELWEISS_DUMP_P1MAP")
```
matthiasneuner: Smelly design!

.gitignore
```diff
@@ -19,3 +19,8 @@ build
 EdelweissFE.egg-info
 docs
 edelweissfe/built_extensions.log
+edelweissfe/linsolve/klu/klu.c
+
+# Investigation/handoff working docs -- local scratch, not project documentation
+PERF_LINSOLVE_INVESTIGATION.md
```
matthiasneuner: Obviously this should not be here !!!!

setup.py
```diff
@@ -208,6 +208,13 @@
 ]
 
 print("Gather the AMGCL interface")
+# Deliberately NOT *arch_flags here, unlike this extension's siblings above: measured on a Xeon
```
matthiasneuner: ??? Crazy?

tests/test_mpc_amgcl_spgemm.py
```diff
@@ -0,0 +1,118 @@
...
+#
+#  This library is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU Lesser General Public
+#  License as published by the Free Software Foundation; either
+#  version 2.1 of the License, or (at your option) any later version.
+#
+#  The full text of the license can be found in the file LICENSE.md at
+#  the top level directory of EdelweissFE.
+#  ---------------------------------------------------------------------
+"""§24 correctness tests for the AMGCL-threaded T^T K T + C expression
```
matthiasneuner: Remove these obvious references to a plan file! Doc should be doc for the code and not "how you got there"

examples/AnchorPryOut/zsymm_free_block.txt
matthiasneuner: Confirmed real gap, not a quick fix: NodeSet has no set-difference operator, and Tie's slave-node claims are only resolved geometrically inside updateConnectivity() (model-setup time), after *nSet parsing -- so this can't currently be expressed in the input language at all, static or not. Filed as issue #92 (general input-language capability, not blockamg-specific) rather than blocking this PR on it.

edelweissfe/linsolve/inexactnewton/inexactnewton.py
matthiasneuner: Confirmed you are right: unlike gmres/mumps/petsclu (whose sibling implementation modules import genuinely optional third-party packages -- pyamg, mumps, petsc4py -- lazily isolated behind the factory), InexactNewtonSolver/_SuperLUFactorization only ever import numpy/scipy (hard dependencies), so the two-file split had no technical justification here. Merged inexactnewton.py into __init__.py; the PARDISO-delegate import stays deferred for the unrelated circular-import reason (config.linsolve imports the registry, which is what resolves this module).

edelweissfe/solvers/nonlinearimplicitstatic.py
matthiasneuner: Agreed, and confirmed: blockamg already reads coordinates/topology lazily from self._model (set by setModel()) rather than from these dump files -- they were pure offline-diagnostic tooling with no reason to be orchestrated by NIST. Moved both dumps (EDELWEISS_DUMP_COORDS/EDELWEISS_DUMP_P1MAP) into LinearSolver.setModel() in linsolve/base.py, where model/dofManager are already available to every registered solver uniformly. Verified end-to-end: files still get written, journal messages now correctly attributed to LinearSolver instead of NIST.

edelweissfe/solvers/nonlinearimplicitstatic.py
matthiasneuner: Same fix as the sibling comment above -- both dump blocks moved to LinearSolver.setModel().

.gitignore
matthiasneuner: Fixed -- removed both entries. A shared .gitignore should not special-case one contributor's personal investigation-doc filenames.

setup.py
matthiasneuner: Real inconsistency, not just a surprising number: every other extension respects EDELWEISSFE_ARCH_FLAGS for per-machine override, but AMGCL silently ignored that mechanism entirely (hardcoding no arch flags unconditionally), so nobody could opt back into -march=native on hardware without the measured Skylake-SP AVX-512-downclock erratum. Fixed to read the same variable with an empty default -- safe by default, same override capability as every sibling extension.

tests/test_mpc_amgcl_spgemm.py
matthiasneuner: Fixed in both test_mpc_amgcl_spgemm.py and test_p1topology.py (same pattern) -- stripped the PERF_LINSOLVE_INVESTIGATION.md phase/task references, kept only the code-relevant rationale.

edelweissfe/linsolve/inexactnewton/__init__.py
matthiasneuner, line 1: Why is there a complete implementation in an __init__.py file ?!?!?!?

matthiasneuner: Why is this even a "linear solver"? ???

edelweissfe/config/registry.py
```diff
@@ -385,6 +385,19 @@ def _addBuiltins(category: str, attrName: str, moduleNames: list[str], packageDo
     "mumps",
     "gmres",
     "amgcl",
+    # not a direct solver in its own right: it preconditions GMRES with a lagged exact LU (from one
```
matthiasneuner: Why is this explained in registry ????? This should be docstrings and in the sphinx doc!!

edelweissfe/linsolve/nullspace.py
matthiasneuner, line 1: Why is this part of blockamg? Seems like something solver independent???

edelweissfe/solvers/nonlinearimplicitstatic.py
```diff
@@ -247,8 +287,35 @@ def solveStep(
...
+                                fieldIndices.start,
+                                fieldIndices.stop,
+                            ),
+                            self.identification,
+                            0,
+                        )
+
                     self.journal.printSeperationLine()
 
+                    # Hand every linear solver the live model and DOF manager -- the one interface
```
matthiasneuner: Why sucha verbose comment???

doc/source/documentation/linsolvers.rst
```diff
@@ -59,3 +71,207 @@ Choose the options for the linsolver (in this case ``gmres``) in an extra file:
 	},
 	"linsolveopts": {"maxiter": 1, "restart": 1500}
 	}
+
+
+The ``inexactnewton`` solver
```
matthiasneuner: This PR introduces many noew solver modules, but it is not explained how they are related to each other

edelweissfe/config/registry.py
```diff
@@ -385,6 +385,19 @@ def _addBuiltins(category: str, attrName: str, moduleNames: list[str], packageDo
...
+    # of the direct solvers above), reused across several Newton iterations under an Eisenstat--Walker
+    # forcing sequence -- a modified-Newton-Krylov scheme that turns most direct solves into a handful
+    # of back-substitutions on the large coupled fracture models where the factorization dominates.
+    "inexactnewton",
```
matthiasneuner: Is this a "linear solver"?

edelweissfe/config/registry.py
```diff
@@ -385,6 +385,19 @@ def _addBuiltins(category: str, attrName: str, moduleNames: list[str], packageDo
...
+    # a field-split block-AMG solver for large coupled multi-field systems: an AMG hierarchy per field
+    # (elasticity displacement, scalar damage) combined by a block Gauss-Seidel sweep to precondition
+    # an outer GMRES. The O(n)-memory route to sizes a direct factorization cannot reach.
+    "blockamg",
+    # not a solver in its own right: it dumps the equation systems it is handed and delegates the
+    # actual solve to one of the above, so linear-solver variants can be compared offline on one
+    # authentic sequence of matrices instead of by rerunning the simulation.
+    "matrixdump",
```
matthiasneuner, line 383: So how is it used????

.gitignore
matthiasneuner, line 1: do not modify .gitignore!

matthiasneuner: Fixed -- .gitignore is now byte-identical to next_v26.11, no modifications at all. (The generated klu.c ignore went with it; that file is simply left untracked.)

edelweissfe/config/registry.py
matthiasneuner: Agreed. Removed the per-solver explanatory comments from the linsolver table -- what each solver does belongs in its own module docstring and the Sphinx docs, not in the registry that merely maps names to dotted strings. The remaining comment just points at both.

edelweissfe/linsolve/nullspace.py
matthiasneuner: You are right, it is solver-independent: both functions are pure geometry/linear algebra over a FieldBlock, its equilibration scaling and its node coordinates, with no solver-instance state. Moved to edelweissfe/linsolve/nullspace.py (next to base.py) and rewrote the docstring accordingly -- blockamg/ptwogrid are now just its first two consumers rather than its owners.

edelweissfe/solvers/nonlinearimplicitstatic.py
matthiasneuner: Fixed -- condensed from 10 lines to 3.

edelweissfe/linsolve/inexactnewton/__init__.py
matthiasneuner: Fair -- that was my doing in the previous round, over-correcting your earlier 'why is this a separate file?' comment by collapsing everything into __init__.py. Now split properly: __init__.py is a one-line re-export of createSolver, and inexactnewton.py holds the implementation *and* its factory together. That satisfies both objections -- no implementation in __init__.py, and the factory is not separated from the class it constructs.

doc/source/documentation/linsolvers.rst
matthiasneuner: Added a 'How the available solvers relate' section to linsolvers.rst, grouping the twelve entries into direct solvers, iterative solvers (gmres/amgcl/blockamg), the inexactnewton hybrid, and the matrixdump diagnostic wrapper -- with what distinguishes each group and when to reach for it.

edelweissfe/config/registry.py
matthiasneuner: Documented in the new overview section: matrixdump writes every (A, b) pair it is handed to disk and delegates the actual solve to a real solver (its 'delegate' option), so the run proceeds normally. The point is to compare solver variants offline on ONE authentic matrix sequence via scripts/benchmark_linsolve.py, instead of rerunning the simulation per variant (slow, and not like-for-like, since a variant that changes the Newton iterates changes the sequence being compared). Investigation tool, never production.

edelweissfe/linsolve/inexactnewton/__init__.py
matthiasneuner: Dropped inexactnewton from this PR entirely. Three verified reasons, not just the framing one:

1. Nothing used it -- only its own regression test. blockamg's single mention was a docstring cross-reference, not code.
2. It duplicated shipped code: blockamg independently carries the same Eisenstat-Walker forcing sequence and residual-jump increment detection. The increment-detection expression is byte-identical between the two files, as are the EW core, safeguard, clamp, and the whole surrounding state. That duplication had already bitten -- the zero-previous-residual guard existed in blockamg's copy but was missing from inexactnewton's, which is exactly the NaN bug flagged earlier in this review.
3. Your framing objection was right: it is a modified-Newton-Krylov *nonlinear* strategy wearing a linear solver's interface, inferring increment boundaries from a 4x jump in ||b|| because it has no access to the Newton loop -- guessing what the nonlinear solver knows for certain.

The measured lagged-LU result stands as a finding and is still reachable offline via benchmark_linsolve.py's 'lagged' subcommand (which is why PardisoSolver's phase-split factorize()/solveFactorized() API is kept -- that script is its real consumer). If it comes back it should be with a real interface where the nonlinear solver tells it about increments, and the shared EW logic factored out once rather than copied a third time.

matrixdump is kept: unlike inexactnewton it is complete, has a real in-tree consumer (scripts/benchmark_linsolve.py reads its manifest.jsonl format), and is the harness this PR's own measurements rest on. Its usage is now documented in the new overview section.

setup.py
```diff
@@ -1,301 +1,316 @@
...
+            "pardiso",
+        ],
+        language="c++",
+        extra_link_args=["-fopenmp", "-lgfortran", "-lpthread", "-lm"],
+        optional=True,
+    )
+]
+
+print("Gather the AMGCL interface")
+# Defaults to no arch flags here, unlike this extension's siblings above (still reads
```
matthiasneuner: What's this verbose bullshit comment here?

doc/source/documentation/blockamgtheory.rst
matthiasneuner, line 1: This page should be nested under "linear solvers"

setup.py
matthiasneuner: Fixed -- cut from nine lines to three, keeping only the conclusion and the opt-in mechanism:

    # No arch flags by default: -march=native measured ~40% SLOWER here on Skylake-SP, where AMGCL's
    # sustained 512-bit inner loops trigger that generation's package-wide AVX-512 downclock. Set
    # EDELWEISSFE_ARCH_FLAGS explicitly to opt in.

doc/source/documentation/blockamgtheory.rst
matthiasneuner: Fixed -- it was a sibling of linsolvers in the Documentation toctree; now a child. Removed it from the parent toctree and gave linsolvers.rst its own toctree containing it.

Verified rather than assumed: the page renders on the linsolvers page as 'Field-split block AMG: theory and configuration', Sphinx raises no 'not included in any toctree' warning (it would if the nesting were broken), and a matched-baseline build with PYTHONPATH pinned to the worktree gives a byte-identical warning set, so the change adds none.

One honest caveat: the top-level Documentation index does not visually expand the child, because that toctree's ':maxdepth: 2' budget is consumed by each page's own section headings. The document hierarchy is correct; making the global nav expand it would mean raising maxdepth for every page, which I did not want to do unilaterally. Say the word if you want that.

matthiasneuner: Good catch, and my earlier 'it is accessible' answer glossed over this. The toctree was appended at the *end* of linsolvers.rst, and in reStructuredText that places it inside the last section -- so it rendered as a child of 'The matrixdump diagnostic solver', which is exactly what you saw.

Moved it into the blockamg section, where the sentence introducing it already sat. Verified in the built HTML by section offsets rather than by eye: the sidebar now reads 'The blockamg solver' -> 'Field-split block AMG: theory and configuration' (toctree-l3 -> l4). The remaining occurrence under matrixdump is just the theme's footer 'Next' button, which always sits at the page bottom.

edelweissfe/linsolve/blockamg/blockamg.py
```diff
@@ -0,0 +1,1173 @@
...
+                        with performancetiming.timeit("nullspace construction"):
+                            # Coordinates come from self._model (set by setModel), read fresh every
+                            # rebuild rather than pushed in ahead of time.
+                            coords = self._getNodeCoordinates(block.name)
+                            if coords is not None:
+                                # Full rigid-body basis (translations + rotations): measured ~28-31%
+                                # fewer isolated outer iterations than translations alone on two real
+                                # captured systems, and robust to both thread count and power_iters
+                                # where translations-only is sensitive to both.
+                                nullspace = rigidBodyNullspace(block, coords, dinv[slices[i]])
```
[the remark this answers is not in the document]
matthiasneuner: Confirmed and fixed. The flag was checked at one call site (the experimental p-two-grid path) and not the other (the default single-level path) -- and the comment at the ungated site claimed parity with 'the non-p1Map path below', which was precisely the path missing the check. Both call sites are now gated; verified there are zero ungated `_getNodeCoordinates(block.name)` sites left. Worth noting this silently invalidated the rigid-body-vs-translations comparison the docs' ~28-31% figure comes from, since this is the knob you would use to reproduce it.

edelweissfe/linsolve/blockamg/blockamg.py
```diff
@@ -0,0 +1,1173 @@
...
+        field = self._model.nodeFields.get(fieldName)
+        if field is None:
+            return None
+        return np.array([node.coordinates for node in field.nodes], dtype=float)
+
+    def _getP1Map(self, fieldName: str):
+        """This field's P1 corner/midside topology map, computed lazily on first need and
+        cached for the lifetime of this instance (never refreshed even across AMR -- matching this
+        opt-in p-multigrid path's original behaviour, unvalidated for a mesh that changes after the
+        map was built and not a focus of this refactor).
```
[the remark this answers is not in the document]
matthiasneuner: Confirmed and fixed. blockamg did not override setModel, and `_getP1Map`'s `if fieldName not in self._p1Maps` guard meant a lazily built map was never recomputed. Added a setModel override that drops lazily computed maps while preserving explicitly supplied offline ones (tracked separately in `_lazyP1MapNames`, since both previously lived in the same dict). Verified with a stub model: stale lazy map dropped, explicit map preserved, base-class field-structure bookkeeping still runs.

edelweissfe/linsolve/blockamg/blockamg.py
```diff
@@ -0,0 +1,1173 @@
...
+        # Recorded *after* the dump above, so a trigger's own context window (from
+        # _recentSolveHistory) never includes itself -- only solves strictly preceding it. Guarded on
+        # dumpOnDegradationContextSolves so a run with the feature off never holds onto extra matrices.
+        if self._dumpOnDegradationDir is not None and self._dumpOnDegradationContextSolves > 0:
+            self._recentSolveHistory.append(
+                {
+                    "solveCount": self._solveCount,
+                    "A": A,
+                    "b": b,
+                    "blocks": blocks,
```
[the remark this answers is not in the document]
matthiasneuner: Confirmed and fixed -- now `A.copy()` / `np.array(b, copy=True)` on enqueue. The premise checks out in the solver's own words: nonlinearimplicitstatic.py documents that 'K is self.csrGenerator's own persistent CSR matrix, returned by reference and reused', and the residual is updated in place via `R[:] = -P`. So every history entry aliased the trigger-time values, which is the exact opposite of the authentic preceding-solve sequence the window exists for.

edelweissfe/linsolve/pardiso/pardiso.pyx
```diff
@@ -254,19 +270,121 @@ cdef class PardisoSolver:
 
         # numerical factorization
         phase = 22
-        pardiso(self.pt, &self.maxfct, &self.mnum, &self.mtype, &phase,
-                &self.rows, &data[0], &self.indptrFortran[0], &self.indicesFortran[0], &idum, &nRhs,
-                &self.iparm[0], &self.msglvl, &ddum, &ddum, &error)
+        with performancetiming.timeit("pardiso phase 22 (numeric factorization)"):
+            pardiso(self.pt, &self.maxfct, &self.mnum, &self.mtype, &phase,
+                    &self.rows, &data[0], &self.indptrFortran[0], &self.indicesFortran[0], &idum, &nRhs,
+                    &self.iparm[0], &self.msglvl, &ddum, &ddum, &error)
```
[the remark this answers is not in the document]
matthiasneuner: Confirmed and fixed. `__call__` never touched hasNumericFactorization/currentData, so a prior factorize() left them set and a later solveFactorized() paired the new factors with the old matrix's values. Demonstrated before/after on diag(1,2,3) then a one-shot on diag(4,5,6): it previously returned a silently wrong vector, and now raises 'no numeric factorization available; call factorize() first'. (Your 'also on line 340' pointer -- that site is inside factorize() itself, where setting the state is correct; there are only two phase-22 call sites and the other one was the bug.)

tests/test_mpc_amgcl_spgemm.py
```diff
@@ -0,0 +1,117 @@
...
+The general MPC-construction tests (chained records, zero slave DOFs) are ported here; the
+scatter-map-specific ones have no equivalent in the AMGCL path, which has no cache to corrupt.
+"""
+
+import os
+
+import numpy as np
+import scipy.sparse as sp
+
+from edelweissfe.numerics.mpctransformation import MultiPointConstraintTransformation
```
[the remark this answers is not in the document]
matthiasneuner: Confirmed and fixed -- added a module-level `pytest.importorskip('edelweissfe.linsolve.amgcl.amgcl')`. All four tests drive `useAmgclSpgemm=True` and setup.py builds that extension as optional, so a SciPy-only install was erroring rather than skipping. One correction: the 'integration test's stated optional-extension behavior' you cite as precedent does not exist -- test_blockamg_forcing_tolerance.py has no gate either. The core point stands regardless.

scripts/block_amg_prototype.py
```diff
@@ -0,0 +1,183 @@
...
+And = A[d:, :d].tocsr()
+Ann = A[d:, d:].tocsr()
+
+# Rigid-body translations for the displacement block: 1 on each component, from the DOF layout alone.
+Bdisp = np.zeros((d, 3))
+Bdisp[np.arange(d), np.arange(d) % 3] = 1.0
+# Under symmetric scaling the near null-space transforms as B -> D^1/2 B (so that A_hat (D^1/2 v) =
+# D^-1/2 A v ~ 0). Without this the AMG is handed the wrong near-null-space on the scaled operator.
+if scale == "on":
+    Bdisp = Bdisp * np.sqrt(np.abs(Add.diagonal()))[:, None]
```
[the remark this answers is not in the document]
matthiasneuner: Confirmed and fixed. Verified the sequence: `A` is equilibrated at line 75, `Add` is sliced from the already-equilibrated `A` at line 88, so scaling by `sqrt(|Add.diagonal()|)` was multiplying by approximately one -- a near no-op, as you say. Now divides by `dinv`, matching the production path (`edelweissfe.linsolve.nullspace`: `B / blockDinv[:, None]`).

edelweissfe/solvers/nonlinearimplicitstatic.py
```diff
@@ -191,6 +228,9 @@ def solveStep(
             if "linsolver" in self.options
             else getDefaultLinSolver()
         )
+        # Every registered linsolver inherits LinearSolver's setJournal() (a safe no-op-ish default for
+        # solvers that do not log), so this is unconditional -- no isinstance check needed.
+        self.linSolver.setJournal(self.journal)
```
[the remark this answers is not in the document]
matthiasneuner: Premise right, conclusion wrong -- so I did not apply the suggested fix. NEST indeed calls neither setJournal() nor setModel(). But it also never *invokes* its linear solver: the only occurrence of `linSolver` in nonlinearexplicitstatic.py was the assignment, and NIST's three uses all sit inside `solveStep`, which NEST overrides. So `BlockAMGSolver.__call__` is unreachable under NEST and nothing raises.

The real defect was that NEST built a linear solver it never used. Rather than mirroring the initialization calls onto a solver that is never called, I removed the dead construction, along with the `linsolver`/`linsolverConfigFile` schema fields and defaults -- an explicit solver should not advertise options it cannot honour (that would be the same silent-ignore problem as the useRigidBodyNullspace finding). `pruneCondensedMatrixZeros` is kept, since NEST replaces SolverSpecificOptions wholesale and reuses NIST's Dirichlet path which reads it.

edelweissfe/config/registry.py
```diff
@@ -385,6 +379,8 @@ def _addBuiltins(category: str, attrName: str, moduleNames: list[str], packageDo
     "mumps",
     "gmres",
     "amgcl",
+    "blockamg",
+    "matrixdump",
```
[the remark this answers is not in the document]
matthiasneuner: Half right, and the actionable half is now done. The removal was deliberate (see the discussion thread on inexactnewton) -- it was unused, duplicated blockamg's forcing/increment-detection logic byte-for-byte, and was a nonlinear strategy in a linear solver's clothing. But you are correct that the PR metadata was stale: the description said three solvers / 12 entries and the title still said 'inexact-Newton solver'. Both updated -- two new linsolvers, registry total 11 -- with a 'Scope change during review' section recording why.

edelweissfe/solvers/nonlinearimplicitstatic.py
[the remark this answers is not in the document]
matthiasneuner: Correction: you were right and I was wrong. CI caught it as `'NEST' object has no attribute 'linSolver'`.

My reasoning was faulty in a specific way: I grepped the two solver subclasses for `self.linSolver`, found only the assignment in NEST, and concluded the solver was never used. The dereference is in neither file -- NEST's own `solveIncrement` calls the inherited `NonlinearSolverBase.linearSolve()` (nonlinearsolverbase.py:313), which is what reads `self.linSolver`. I searched for the attribute name and missed the indirection through the base-class helper. NEST solves a linear system per Runge-Kutta stage; 'explicit' refers to the time integration, not to avoiding a linear solve.

Reverted the removal and applied your suggested fix instead: `setJournal()` and `setModel()` after NEST constructs its DOF manager and linear solver. NEST builds its equation system once rather than rebuilding on connectivity changes, so one call suffices.

Verified with a pure-Python NEST case, since the Marmot-backed NEST/NESTParallel fixtures cannot run on this machine: the default path completes, and `linsolver=blockamg` completes while performing 32 real linear solves with per-field hierarchy builds -- previously the exact configuration that raised.

edelweissfe/solvers/nonlinearexplicitstatic.py
```diff
@@ -298,6 +310,14 @@ def solveStep(
             with open(linsolverOptions, "r") as f:
                 linsolverOptionDict = json.load(f)
         self.linSolver = getLinSolverByName(self.options.get("linsolver", "default"), linsolverOptionDict)
```
alexdummer, line 312: **Correctness:** NEST constructs its `linSolver` here but never calls `setJournal()`/`setModel()` on it, unlike NIST. If a step configures `linsolver: blockamg` (or `matrixdump` wrapping blockamg) for NEST, `BlockAMGSolver._fieldStructure` is left unset (only `LinearSolver.setModel()` populates it), so the first linear solve raises `RuntimeError('blockamg: field structure not set...')` with no earlier validation to catch it.

edelweissfe/linsolve/blockamg/blockamg.py
```diff
@@ -0,0 +1,1199 @@
...
+        )
+        # A's sparsity pattern churns between Newton iterations on this class of problem (condensed
+        # contact/tie systems, whose active constraint set changes every Newton iteration) -- a
+        # hierarchy built for a different pattern is not just "a bit stale", it can be a drastically
+        # worse preconditioner (measured: 494 vs. 94 outer iterations on one such transition, an
+        # outright wall-clock regression, not a graceful few-extra-iterations degradation). ``nnz`` is a
+        # cheap, free (O(1)) proxy for "the pattern changed" -- not exact (two different patterns could
+        # coincidentally share a total nnz), but it caught the one measured failure case and errs in the
+        # safe direction (an unnecessary refresh costs time, never correctness).
+        patternChanged = self._lastNnz is not None and A.nnz != self._lastNnz
```
alexdummer, line 813: **Correctness:** the diagonal equilibration `dinv = 1/sqrt(|diag(A)|)` has no floor against a zero diagonal entry. A Lagrange-multiplier DOF from a constraint (`equalvaluelagrangian.py` never writes `K[multiplier, multiplier]`, leaving it structurally zero) produces `dinv=inf`, corrupting `As`/`bs` with inf/NaN before the AMG hierarchy build and outer solve — a silent NaN Newton correction instead of a controlled error. Other divisions in this file are floored at 1e-300; this one isn't.

edelweissfe/linsolve/matrixdump/matrixdump.py
```diff
@@ -0,0 +1,227 @@
...
+
+        self._instanceOrdinal = MatrixDumpSolver._instancesCreated
+        MatrixDumpSolver._instancesCreated += 1
+
+        self._solveCounter = 0
+
+        os.makedirs(self._directory, exist_ok=True)
+        self._manifestPath = os.path.join(self._directory, "manifest.jsonl")
+
+    def setJournal(self, journal) -> None:
```
alexdummer, line 126: **Correctness:** `MatrixDumpSolver` forwards `setJournal`/`setFieldStructure`/`setModel` to its delegate but not `factorize()`/`solveFactorized()`, and has no `__getattr__` fallback. Wrapping a `PardisoLinearSolver` delegate here (to capture matrices while also using the lagged-LU factorize/solveFactorized phase-split path that `scripts/benchmark_linsolve.py`'s `lagged` subcommand relies on) raises `AttributeError` as soon as `factorize()`/`solveFactorized()` is called on the wrapper.

edelweissfe/linsolve/amgcl/amgcl-wrapper.hpp
```diff
@@ -79,7 +169,697 @@ class LinearSolver {
...
+
+    if ( !solver_ || n != cached_n || nnz != cached_nnz || !std::equal( ptr, ptr + n + 1, cached_ptr_.begin() ) ||
+         !std::equal( col, col + nnz, cached_col_.begin() ) ) {
+      solver_.reset( new Solver( amgcl::adapter::block_matrix< BlockType >( A ), prm ) );
+      cached_n   = n;
+      cached_nnz = nnz;
+      cached_ptr_.assign( ptr, ptr + n + 1 );
+      cached_col_.assign( col, col + nnz );
+    }
+    else {
```
alexdummer: **Correctness:** in `LinearSolverBlockT::solve()`, the "pattern unchanged" branch still unconditionally rebuilds the AMG hierarchy via `solver_.reset(new Solver(...))`, identical to the "pattern changed" branch — the caching guard is a no-op. Reachable via `linsolver: amgcl` with `backendBlockSize: 2` or `3` (wired from `amgcl.pyx`'s `PyAMGCLSolver.solve()`, exercised by `testfiles/marmot/AMGCL/test.inp`) — every call pays a full hierarchy rebuild even when the sparsity pattern hasn't changed. The scalar `LinearSolverT::solve()` has the identical defect.

edelweissfe/linsolve/blockamg/ptwogrid.py
```diff
@@ -0,0 +1,314 @@
...
+        nNodes = len(self._isCorner)
+        if n % nNodes != 0:
+            raise ValueError(
+                "ptwogrid: block size {:} is not a multiple of the topology map's node count {:} -- "
+                "the map does not match this field.".format(n, nNodes)
+            )
+        nDim = n // nNodes
+        P_dof = sp.kron(self._P_node, sp.identity(nDim), format="csr")
+
+        dirichletMaskBool = np.diff(A.indptr) == 1
```
alexdummer, line 249: **Correctness:** Dirichlet rows are identified purely by nnz-per-row==1 (`np.diff(A.indptr)==1`), which isn't guaranteed to hold. `dirichlet.pyx` zeroes off-diagonals but keeps them as explicitly-stored zeros; they're only pruned to nnz==1 when an MPC/tie transformation is active *and* `pruneCondensedMatrixZeros` is set (via `K.eliminate_zeros()` in `nonlinearimplicitstatic.py`). On a plain-Dirichlet-BC model with no MPC/tie constraints, Dirichlet rows never get pruned, so this heuristic silently fails to identify them — `applyPreconditioner()` would then mistreat a genuinely eliminated row as a free unknown, or vice versa for a structurally isolated free DOF.

edelweissfe/linsolve/pardiso/__init__.py
```diff
@@ -35,6 +35,36 @@
...
+    contracts, not just the ``LinearSolver`` one.
+    """
+
+    def __init__(self, delegate):
+        self._delegate = delegate
+
+    def __call__(self, A, b):
+        return self._delegate(A, b)
+
+    def factorize(self, A):
```
alexdummer, line 62: **Simplification/reuse:** `PardisoLinearSolver` adds `factorize()`/`solveFactorized()` as a phase-split capability entirely outside the shared `LinearSolver` base contract (`base.py` defines only `__call__`/`setJournal`/`setModel`/`setFieldStructure`). Any generic `LinearSolver` consumer or wrapper (the `matrixdump.py` finding above is the concrete instance) can't discover or safely proxy this capability without an out-of-band isinstance/duck-type check. Worth adding to the base contract (even as an optional/default no-op) so a future solver or wrapper doesn't repeat the same silent interface gap.

edelweissfe/linsolve/blockamg/blockamg.py
```diff
@@ -0,0 +1,1199 @@
...
+            # linear-elastic, prior iterate) makes the ratio undefined rather than just
+            # uninformative -- fall back the same way as the no-history case.
+            return self._etaMax
+
+        ratio = residualNorm / self._lastResidualNorm
+        eta = self._ewGamma * ratio**self._ewAlpha
+
+        safeguard = self._ewGamma * self._lastEta**self._ewAlpha
+        if safeguard > 0.1:
+            eta = max(eta, safeguard)
```
alexdummer, line 537: **Efficiency:** `_getNodeCoordinates()` rebuilds a field's coordinate array via a pure-Python list comprehension on every AMG hierarchy refresh, even though node coordinates never change. `mustRefresh` (per this file's own comments) fires essentially every Newton iteration on contact/tie models where the sparsity pattern churns, so this O(nNodes) Python loop (called at lines 867 and 893) reruns on every linear solve instead of once per analysis. Worth caching per field.

edelweissfe/linsolve/blockamg/ptwogrid.py
```diff
@@ -0,0 +1,314 @@
...
+
+    def smooth(x, rhs, sweeps):
+        for _ in range(sweeps):
+            smoother.applyStep(x, rhs)
+        return x
+
+    return smooth, smoother.residual
+
+
+def buildNodeLevelP(isCorner: np.ndarray, edgeEndpoints: np.ndarray) -> sp.csr_matrix:
```
alexdummer, line 130: **Efficiency:** `buildNodeLevelP()` rebuilds the P1 restriction operator via an O(nNodes) pure-Python loop every time `PTwoGridPreconditioner` is constructed fresh inside `blockamg.py`'s `mustRefresh` block, even though it depends only on already-cached topology (`isCorner`/`edgeEndpoints`). The topology-level cache (`_p1Maps` in blockamg.py) sits one layer above this function and does nothing to stop this from re-deriving the identical restriction matrix on every refresh — repeated, non-vectorized recomputation on every Newton iteration for churning contact models.

edelweissfe/linsolve/blockamg/blockamg.py
```diff
@@ -0,0 +1,1199 @@
...
+        """The field blocks tiling ``[0, n)``, in DOF order, with any trailing DOFs not covered by a
+        node field (e.g. scalar variables) folded into a final scalar block."""
+
+        if self._fieldStructure is None:
+            raise RuntimeError(
+                "blockamg: field structure not set. It is derived from setModel() (or, at the lower "
+                "level, setFieldStructure()); this solver must be driven by a caller that calls one of "
+                "those."
+            )
+        blocks = sorted(self._fieldStructure, key=lambda field: field.start)
```
alexdummer, line 616: **Simplification/reuse:** three independent hand-rolled implementations of the same sparse-array dump-writer pattern (uncompressed save, process-wide write-budget counter, manifest, progress log) exist here (`_dumpOneSystem`), in `matrixdump.py`'s `_dump`, and in `mpctransformation.py`'s `_dumpForOfflineProbe`, with no shared helper. This file's own docstring says its format deliberately matches `MatrixDumpSolver`'s "so the same offline tooling can read either," yet reimplements the writer from scratch — a future format change (checksum, compression, counter-decrement fix) applied to only one of the three breaks shared offline replay tooling for that variant, and the three write-budget counters can drift independently.

alexdummer (commented):
Automated review pass (correctness + reuse/simplification/efficiency), 10 findings ranked most-severe first as inline comments below. One additional candidate (base.py's setModel unconditionally calling dump helpers) was checked and refuted — dump decision and computation are correctly gated behind EDELWEISS_DUMP_* env vars, no comment posted for it.

edelweissfe/linsolve/blockamg/blockamg.py
```diff
@@ -0,0 +1,1199 @@
...
+            x = [np.zeros(sizes[i]) for i in range(nFields)]
+            for _ in range(self._sweeps):
+                sweepOnce(range(nFields), residual, x)
+                if self._symmetric:
+                    sweepOnce(range(nFields - 1, -1, -1), residual, x)
+            return np.concatenate(x)
+
+        preconditioner = LinearOperator((n, n), matvec=blockGaussSeidel, dtype=As.dtype)
+
+        if self._outerTol is not None:
```
alexdummer, line 1005: **Simplification/reuse:** the first-pass outer-solve dispatch (scipy-vs-amgcl_lgmres branch, lines 965-998) is duplicated almost verbatim in the true-residual continuation loop (lines 1049-1074), differing only in variable naming and `x0=z`. The info/iteration-extraction logic is already separately copy-pasted between the two blocks, so a fix applied to one path but not the other is an easy, silent bug. Worth factoring into a shared helper.

edelweissfe/solvers/nonlinearexplicitstatic.py
matthiasneuner: Already fixed, in de45a4dd -- setJournal()/setModel() are at lines 319-320, seven lines below the line this anchors to. (Copilot raised the same thing; I had wrongly removed NEST's linear solver entirely on the theory that it was never used, CI caught it as "'NEST' object has no attribute 'linSolver'", and the fix went in. NEST does solve a linear system per Runge-Kutta stage, via the inherited NonlinearSolverBase.linearSolve().) Verified with a pure-Python NEST case: linsolver=blockamg completes, 32 real linear solves with per-field hierarchy builds.

edelweissfe/linsolve/blockamg/blockamg.py
matthiasneuner: Confirmed and fixed in 2ae2999e -- verified every link: equalvaluelagrangian.py:198-226 writes the multiplier's off-diagonal coupling and never K[multiplier,multiplier]; blockamg accepts such a model (it appends a FieldBlock('scalar variables', ...) for exactly those dofs); and numerically dinv -> inf makes both As and bs non-finite.

One correction though: matching the other divisions' 1e-300 floor would not fix it. Measured -- a 1e-300 floor turns inf into ~1e150 and max|As| into ~3.2e149, finite and equally useless for the hierarchy build. A structurally zero diagonal means the row cannot be equilibrated at all, which for a saddle-point system is a statement about what this field-split AMG scheme can handle. So it now raises, naming the offending dof count and first index, rather than limping on. Previously it surfaced as DivergingSolution from NIST's NaN check, which misreads as a physics problem.

edelweissfe/linsolve/matrixdump/matrixdump.py
matthiasneuner: Confirmed and fixed in 2ae2999e. Verified: the class defined setJournal/setFieldStructure/setModel/__call__ and neither factorize/solveFactorized nor __getattr__.

Fixed together with your base-contract point below rather than with a capability probe, since this project's conventions rule out hasattr/getattr sniffing: factorize()/solveFactorized() are now declared on LinearSolver, so matrixdump forwards them unconditionally. Verified end-to-end -- matrixdump wrapping a pardiso delegate now does factorize() + solveFactorized() correctly (returns the exact solution) and writes A/b/manifest.jsonl on the way through.

edelweissfe/linsolve/pardiso/__init__.py
matthiasneuner: Agreed, and done in 2ae2999e -- this turned out to be the right way to fix the matrixdump finding too. factorize()/solveFactorized() are now part of the LinearSolver contract in base.py, defaulting to a raise that names the solver which cannot do it, so a wrapper proxies them polymorphically and no caller needs an isinstance/duck-type check. That also keeps it consistent with the project's no-hasattr convention.

edelweissfe/linsolve/amgcl/amgcl-wrapper.hpp
matthiasneuner: Confirmed on both variants -- the reset sits inside the else at line 169 (scalar) and 311 (block), so neither guard could ever skip anything. Fixed in 2ae2999e, but not by making the cache work, and that distinction matters:

the unconditional rebuild is load-bearing. amgcl's make_solver binds the matrix *values* at construction, and ( *solver_ )( rhs, x ) takes no matrix argument in either overload -- I checked both call sites. So a solver kept across calls solves the system it was built for. Values change on every Newton iteration even when the pattern does not, so honouring an unchanged pattern would silently return the previous iteration's answer. Your 'every call pays a full hierarchy rebuild' is exactly right as a statement of fact; it just cannot be avoided through this API.

What was actually dead was the bookkeeping and the branch structure implying an optimisation that cannot exist here. Removed both (cached_n/cached_nnz/cached_ptr_/cached_col_ and their writes, which were only ever read inside those dead branches) and documented the constraint in their place. Verified: the amgcl extension recompiles, clang-format clean, repeated solves stay exact (2.8e-14) and a values change is picked up correctly rather than reusing a stale hierarchy.

edelweissfe/linsolve/blockamg/blockamg.py
matthiasneuner: Confirmed and fixed in 2ae2999e. Now cached per field and dropped in setModel() -- the one point where the mesh can actually differ -- alongside the P1 map cache. Verified the second call returns the same array object and that setModel() invalidates it.

edelweissfe/linsolve/blockamg/ptwogrid.py
matthiasneuner: Confirmed and fixed in 2ae2999e. PTwoGridPreconditioner now takes an optional pre-built pNode and exposes it as a property; BlockAMGSolver caches it per field next to the coordinate and P1-map caches, invalidated together in setModel(). So the O(nNodes) construction happens once per mesh rather than once per hierarchy refresh.

edelweissfe/linsolve/blockamg/blockamg.py
matthiasneuner: Confirmed -- three independent writers, and blockamg's own docstring says its format deliberately matches MatrixDumpSolver's 'so the same offline tooling can read either' while reimplementing it. Filed as issue #94 rather than grown into this PR; it wants one shared helper, with scripts/benchmark_linsolve.py as the consumer that has to keep reading whatever they emit.

edelweissfe/linsolve/blockamg/blockamg.py
matthiasneuner: Confirmed, and it has already produced a live discrepancy worth recording: the two arms judge convergence against different tolerances. The first pass solves to firstPassEta, but the amgcl_lgmres arm sets info = 0 if lastError <= eta -- comparing against eta, not the tolerance that pass was given. With gapCompensatedTolerance=true (firstPassEta < eta) a solve can report converged against a tolerance it was never asked to meet, while the scipy arm judges against firstPassEta as scipy computes it. Exactly your 'a fix applied to one path but not the other', already realised. Filed as issue #94 with that as the concrete argument for factoring the dispatch.

edelweissfe/linsolve/blockamg/ptwogrid.py
matthiasneuner: Confirmed -- premise verified in the codebase's own words. dirichlet.pyx states at the point it zeroes them: 'the zeroed off-diagonals are deliberately kept as explicitly stored zeros', and the only eliminate_zeros() is nonlinearimplicitstatic.py:814, gated on mpcTransformation is not None and pruneCondensedMatrixZeros. So on a plain-Dirichlet model nnz != 1 and the detection silently fails, as you describe.

Filed as issue #94 rather than fixed here: it is confined to the experimental opt-in p-multigrid path, and the right fix is to pass the Dirichlet DOF indices down from the solver that applied them rather than sharpen a sparsity-pattern heuristic -- more than this PR should take on.

Issue #93, opened 2026-08-26
tests/ is not run by CI, and 5 unit tests are failing on next_v26.11


matthiasneuner:
## Summary

The `tests/` directory (248 pytest unit tests) is **not executed by any CI workflow**, and as a
result 5 of them have been failing on `next_v26.11` unnoticed.

Found while triaging PR #91: I had been reporting these as "pre-existing failures" and was
challenged on it — correctly, since trunk should be green. Verifying properly showed both halves of
the problem: the failures are genuinely pre-existing, *and* nothing guards against them.

## Reproduction

Verified in a clean worktree checked out at `32a9b167` (current `next_v26.11`), built with
`pip install -e .`:

```
$ python -m pytest tests/ -q
5 failed, 237 passed, 11 skipped
```

```
FAILED tests/test_inputlanguage_golden.py::test_inputlanguage_surface_matches_golden
FAILED tests/test_registry.py::test_keyword_category_covers_all_21_top_level_keywords_after_u2b
FAILED tests/test_schemasurface.py::test_module_section_matches_golden_byte_for_byte[edelweissfe.modelmodifiers.adaptivity.hadaptivity]
FAILED tests/test_schemasurface.py::test_module_section_byte_identical_set_is_exactly_the_expected_closed_set
FAILED tests/test_stepoptions.py::test_the_dynamic_options_branch_is_declared_in_exactly_one_place
```

## The CI gap

`.github/workflows/` contains `format.yml`, `run_tests_with_marmot.yml`,
`run_tests_without_marmot.yml` and `sphinx.yml`. Between them they run `pre-commit`,
`run_tests_edelweissfe` on `testfiles/edelweiss-only/` and `testfiles/marmot/`, and `sphinx-build`.

**None of them invokes `pytest`.** `grep -rn pytest .github/workflows/` returns nothing. So the
`.inp`-driven regression suites are covered but the unit-test suite is not, which is how these
five rotted without anyone seeing it.

## The five, with causes

Four are stale assertions/snapshots. One looks like a genuine signal.

1. **`test_registry`** — asserts exactly *21* top-level keywords; there are now **22**. `#58` added
   the `elementProperty` keyword without updating the count (the test name itself still says
   `..._21_top_level_keywords_after_u2b`). Stale count.

2. **`test_inputlanguage_golden`** — the golden snapshot of the input-language surface
   (`printKeywords()` + module documentation) has drifted. The test prints its own remedy:
[...]

PR #95, opened 2026-08-27
feat(topology): plan/apply model-modifier pipeline with a monotonic element allocator

matthiasneuner:
## Summary

Replaces the ad-hoc "iterate modifiers, then iterate constraints" update with an explicit two-phase
topology pipeline:

1. **Model modifiers run to a fixed point inside an open topology window** (`FEModel.updateTopology`),
   split into `plan()` and `apply()`. `plan()` decides; `apply()` mutates and is a pure function of
   `(model, plan)` — so exactly one code path changes the mesh.
2. **Mesh-dependent consumers then catch up once**, on the net change, with the window *closed* — a
   consumer that tries to create or delete an element now raises instead of quietly mutating behind
   the pipeline's back.

Supporting pieces: a monotonic element-number allocator (`reserveElementNumbers`/`createElement`)
replacing per-generator `max()+1` counters and their label-collision risk; reproducible
entity-creation order; a `topologyFingerprint` that makes mesh state checkable; contact-facet
generation promoted to a model modifier of its own; and a pull-based refresh phase replacing the
push observer (`Tie.reconcile` → `Tie.refresh`).

Next in the merge order after #91. Restart lands after this, deliberately.

## How this was built

`feat/topology-pipeline` is 353 commits ahead of trunk but only **39 are its own work** — everything
below sits in the already-merged stack (#64, #68, #89, #91). Diffed from the real fork point
(`perf/linsolve-restart`) and applied with `git apply -3`, per the recipe used for #61/#89/#91,
rather than rebasing 353 commits through dead history.

### Deliberately excluded

| What | Why |
|---|---|
| 2 blockamg commits (`f513e0a4`, `b10d765b`) | Already upstream via #91 under **different hashes and subjects**, so neither `git cherry` nor a subject grep flags them. All four linsolve files in the range are touched only by these two, and trunk's versions are strictly newer (same optimisations + #91's review fixes). |
| Restart serialization | Trunk has no `topologyHistory` persistence and restart lands last. Both conflicts inside `writeRestart`/`readRestart` resolved to trunk. The in-memory side (recording, `replayTopologyHistory`, `restoreDecisionState`) is topology core and **is** included — just not written to disk yet. |
| `tests/test_restart_integration.py` | Drives `*output, type=restart`, an output manager that does not exist on trunk. Belongs with the restart PR. |
| `PLAN_TOPOLOGY_PIPELINE.md` + plan/phase refs | Stripped `"P3"`, `"P1 of PLAN_..."`, `PLAN_TRANSPARENT_AMR.md`, `PLAN_RESTART.md` from code and tests. Docs describe the code, not the route to it. |
| MKL instruction-set pin + `OMP_NUM_THREADS=1` forcing (`61125069`) | **Measured, not assumed** — see below. |

### On the MKL pin

Its own comment cites `AMR_MinMarkedElements` and `AMR_RecoveryError` failing on a Xeon Gold 6140.
The baseline for this work was taken on **exactly that CPU** (AVX-512 confirmed), where **all 18 AMR
tests pass without the pin** — the sensitivity it compensated for was root-caused properly during
#64's post-merge work. The `OMP_NUM_THREADS=1` forcing was removed by that same work as a broken
workaround (it sat *after* the extension import, too late to take effect); reintroducing it would
undo a real fix.

### The one real judgement call

`AMR_ContactRefineShear`'s reference: both sides had regenerated it — trunk after #64's
asymmetric-footprint geometry fix, this branch after its entity-ordering change. Since `test.inp` is
not in the range, the input is trunk's. Regenerating against the transplanted code reproduced
**trunk's** reference to `1.7e-16`, while this branch's differed by `1.8e-1`. Trunk's is kept — which
also shows the entity-ordering change is solution-preserving on trunk's geometry.

## Test plan

Verified on the same machine as the trunk baseline, comparing failures **by name**:

- [x] `testfiles/marmot/`: 114 ran, 2 failures — `IndirectDisplacementControl`, `OutputManagers`, both
      *"latex was not able to process"*, **identical to the trunk baseline. Zero new failures.**
- [x] `testfiles/edelweiss-only/`: 47 ran, 1 failure (`MeshPlot`, same latex cause)
- [x] `pytest tests/`: **286 passed** (trunk: 237 — so +49), 5 failed — the same five already red on
      trunk, see #93
- [x] `pre-commit` clean
- [x] `marmotelement` builds — this required rebuilding Marmot from its `next_v26.11`; the previously
      installed one lacked the `getPropertyNames` API #58 needs, which had been silently *skipping*
      every Marmot-backed test

> Note: #93 (pytest absent from CI) is worth fixing before or alongside this — the pipeline's own
> three test modules live in `tests/`, which no workflow currently runs.


edelweissfe/modelmodifiers/surfacefacets/surfacefacets.py
```diff
@@ -0,0 +1,135 @@
...
+            recipe = model.contactFacetRecipes.get(facetSetName)
+            if recipe is None:
+                continue
+            before = {el.elNumber for el in model.elementSets.get(facetSetName, [])}
+            buildContactFacets(model, *recipe, self._journal)
+            after = {el.elNumber for el in model.elementSets[facetSetName]}
+            change.removedElements |= before - after
+            change.addedElements |= after - before
+            change.changedElementSets.add(facetSetName)
+        return change
```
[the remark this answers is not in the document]
matthiasneuner: Confirmed and fixed in d040b1d8 -- traced the mechanism rather than take it on faith: recordTopologyChange only appends to topologyHistory (the replay record), while notifyModelChanged is what bumps topologyVersion and appends to _changeLog. hadaptivity self-notifies, surfaceFacets did not, so facet regeneration never advanced the version. It went unnoticed because facets only regenerate alongside AMR, whose own notify bumps the version in the same update.

Fixed structurally rather than by adding the missing call. There were two channels for one fact -- return a ModelChange *and* call notifyModelChanged -- and a modifier could satisfy one and forget the other. Now recordTopologyChange records both the history entry and the changeset, and hadaptivity no longer announces its own change: every modifier reports in exactly one way, by returning a ModelChange from apply(). Safe because notifyModelChanged is recording-only (no synchronous callbacks; consumers refresh once afterwards), so the call site does not matter.

One subtlety worth flagging: the call went into recordTopologyChange, not updateTopology. Both callers of apply() -- the live loop and replayTopologyHistory -- share it, so replay keeps recording changesets. Putting it in updateTopology alone would have silently broken replay.

Verified directly: a facet-only change now bumps topologyVersion, lands in the change log and is visible to changesSince, while a None change still does not.

edelweissfe/generators/microstructuregenerator.py
```diff
@@ -321,10 +322,14 @@ def replicateMesh(
...
+                # recycled, so the dict has gaps, and len()+1 could land on a live element and
+                # silently overwrite it.
+                (elNumber,) = model.reserveElementNumbers(1)
+                newEl = elementType(elTypeName, elNumber)
                 nodeList = [model.nodes[nid + 1] for nid in new_el]
                 newEl.setNodes(nodeList)
-                model.elements[len(model.elements) + 1] = newEl
+                model.createElement(newEl)
                 # add element to corresponding element set
                 new_el_ids.append(len(model.elements) - 1)
```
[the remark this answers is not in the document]
matthiasneuner: Confirmed and fixed. Good catch -- it is an internal contradiction: the comment a few lines above explicitly rejects this reasoning for element *creation* ('element numbers are never recycled, so the dict has gaps, and len()+1 could land on a live element') and correctly switches to reserveElementNumbers, but the lookup path kept the discredited assumption.

It passes today only because a freshly built mesh is gap-free, so elNumber coincidentally equals len(model.elements) -- and it breaks the moment a gap exists, which is precisely what this PR's allocator introduces. Now keeps the created element itself instead of a positional guess.

edelweissfe/models/femodel.py
```diff
@@ -82,19 +90,408 @@ def __init__(self, dimension: int):
...
+                if modifier is None:
+                    raise TopologyError(
+                        "the checkpoint records a decision by model modifier {!r}, which this model "
+                        "does not define -- the input file must declare the same modifiers as the run "
+                        "being resumed".format(record.modifier)
+                    )
+                modelChange = modifier.apply(self, modifier.decodePlan(record.plan))
+                replayed = self.recordTopologyChange(
+                    record.roundNumber, record.modifier, modifier, modifier.decodePlan(record.plan), modelChange
+                )
```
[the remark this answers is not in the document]
matthiasneuner: Fixed -- decoded once into a local and passed to both apply() and the record.

examples/AmrRoundProfile/profile_amr_round.py
```diff
@@ -0,0 +1,192 @@
...
+
+    out = subprocess.run(["edelweissfe", "test.inp"], cwd=case, capture_output=True, text=True, check=False).stdout
+
+    timings = {}
+    for stage in STAGES:
+        # the performancetiming table prints "| <stage> | <value> s |", wrapped to a fixed width
+        match = re.search(re.escape(stage) + r"\s*\|\s*([0-9.eE+-]+)\s*s", out)
+        if match:
+            timings[stage] = float(match.group(1))
+    timings["_elements"] = float(len(re.findall(r"^", out)))  # placeholder, replaced below
```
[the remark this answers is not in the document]
matthiasneuner: Confirmed -- verified empirically that re.findall(r'^', out) returns 1 without re.MULTILINE (1 vs 3 on a three-line string), so the fallback was always 1.0. Rather than add the flag, the fallback is now NaN: a line count was never a meaningful element count, and 1.0 made a failed parse look like a plausible scaling summary instead of showing a gap.

matthiasneuner:
An independent review round found a **regression this PR's own verification had missed**, now fixed in `debde78c`.

### The regression: `TieNED` went PASSED → SKIPPED

`ensureSurfaceFacetModifier` registers an implicit `surfaceFacets` model modifier for any `*surface` facet recipe. `validateModelCapabilities` then rejected the model because `supportsModelModifiers` defaults `False` and NED/NEST do not set it — even though NED sets `supportsMPC = True` deliberately, i.e. ties *did* work on explicit dynamics before this PR. The error also blamed h-adaptivity for a model using none.

**Why it hid:** the harness records this as `SKIPPED`, not `FAILED`, so the failure count was unchanged. My "zero new failures" check compared failures by name but never compared *skips* — a PASSED→SKIPPED transition was invisible to it.

**Fix:** `ModelModifierBase.initiatesTopologyChanges = True`, overridden to `False` by the purely reactive `surfaceFacets`. The guard now counts only self-starting modifiers and names them:

> The NEDSolver solver does not run the topology update, so the model modifier(s) ... would never modify the model.

The default is `True`, so a newly added modifier is guarded unless it explicitly opts out. Real h-adaptivity under NED is still refused — verified both directions with new tests.

### The `d040b1d8` contract was untested

The test doubles called `notifyModelChanged` inside `apply()` *and* returned the change (double notify), and deleting the pipeline's notify call left **all 31 tests green**. The doubles now obey the contract, and a new test pins it: with the notify call removed, **5 tests fail** (verified by breaking and restoring).

### Also fixed

- `nodeToDiscreteRigidBodyPenalty`: `__init__` filtered `rpNode` out of the slave nodes, `refresh()` did not — one shared helper now. (The filter is unreachable in practice: `__init__` already raises if a rigid-body node is in the set.)
- `ModelModifierBase.updateModel` removed — dead *and* broken (called `plan`/`apply` outside a topology window, so it would have raised). Stale references in `nonlinearsolverbase.py` and `adaptivitytheory.rst` fixed.
- `replayTopologyHistory` re-stamped `time=self.time` instead of `record.time`, corrupting a replayed history's times; the existing test could not see it (model time 0.0). Fixed with a test that fails on revert.
- Duplicate whole-mesh `elForEid` build per increment; mislabelled `@timeit` on a setup-only method; leftover plan-file/phase references in 8 files.

### Verification, redone properly

Both suites compared against trunk **by name *and* status**, so a skip transition cannot hide again:

| suite | trunk | branch |
|---|---|---|
| `edelweiss-only` | 44P 1F 3S | 44P 1F 3S — identical sets |
| `marmot` | 74P 2F 40S | 74P 2F 40S — identical sets |

`pytest tests/`: 290 passed, plus the same 5 that fail on trunk (#93). Pre-commit clean.

Deferred to the restart PR, with reasons: `RESTART_FORMAT_VERSION = 2` invalidating existing checkpoints for no content change, and `topologyFingerprint()` costing ~0.19 s per applied decision at 64k elements for a replay path unreachable on trunk.

edelweissfe/modelmodifiers/adaptivity/hadaptivity.py
```diff
@@ -502,25 +563,32 @@ def _materialize(self, model: FEModel, records: dict):
         active = set(mesh.active())
         materialized = set(self._eidToEl.keys())
         newValues = {fieldName: {} for fieldName in oldValues}  # interpolated values for new nodes
-        newChildEids = active - materialized
+        # Only children whose parent is already materialised: the batched restart replay can leave
+        # several refinement levels pending at once, and this keeps each pass to one level.
+        newChildEids = {eid for eid in (active - materialized) if mesh.elements[eid]["parent"] in self._eidToEl}
```
alexdummer: **Possible bug: cascaded refinement can permanently orphan grandchildren from `model.elements`.**

`_materialize()`'s parent-membership filter (`newChildEids = {eid for eid in (active - materialized) if mesh.elements[eid]["parent"] in self._eidToEl}`) assumes only one level of new refinement happens per `apply()` call, but `balance_2to1()` can cascade two levels within one call. A grandchild whose intermediate parent was never a materialized leaf is silently and permanently excluded from `model.elements`.

Scenario: a region refined earlier to a high level sits next to a newly-marked coarse region. `refine()` + `balance_2to1()` subdivides a freshly-created child again in the same `apply()` call because a distant neighbour forces it. That grandchild stays 'active' in the octree mirror forever but is never created in `model.elements` — its intermediate parent is never in `self._eidToEl` and can never become a leaf again — permanent silent mesh/model desync in ordinary multi-level AMR, where the old code would have raised a loud `KeyError` instead.

edelweissfe/generators/microstructuregenerator.py
```diff
@@ -314,25 +315,31 @@ def replicateMesh(
...
                 el = all_elements_to_copy[local_el_id]
                 new_el = []
                 for nid in el:
                     k = np.where(associated_nodes_array[:, 0] == nid)[0][0]
                     new_el.append(int(associated_nodes_array[k, 1]))
-                newEl = elementType(elTypeName, len(model.elements) + 1)
+                # len(model.elements) + 1 was not merely unidiomatic: element numbers are never
+                # recycled, so the dict has gaps, and len()+1 could land on a live element and
+                # silently overwrite it.
+                (elNumber,) = model.reserveElementNumbers(1)
```
alexdummer, line 332: **Possible bug: node numbering still uses the ad hoc `len(model.nodes)+1` pattern the PR explicitly rejected for elements — no allocator guard exists for nodes.**

This hunk correctly replaces `len(model.elements) + 1` with `model.reserveElementNumbers(1)` for elements, but a few lines above (line 297: `_node = Node(len(model.nodes) + 1, np.array(new_nodes[-1])); model.nodes[len(model.nodes) + 1] = _node`), node numbering is untouched and still uses the exact pattern this PR is fixing for elements. There's no `FEModel.reserveNodeNumbers`/`createNode` equivalent.

Scenario: two independent modifiers (e.g. two hAdaptivity instances with disjoint `refineElSets`, so `checkModelModifierDomains` — which only checks element-set overlap — passes cleanly) each seed a node registry from the same `max(model.nodes.keys())` snapshot and independently mint 'new' node numbers over subsequent increments. They can collide on the same label; the second modifier's commit code silently treats the first modifier's node as its own, wiring the wrong node/coordinates into its new child's connectivity with no error — silent wrong-topology corruption, not a crash.

edelweissfe/models/femodel.py
```diff
@@ -82,19 +90,439 @@ def __init__(self, dimension: int):
...
+            roundNumber = 0
+            while True:
+                roundNumber += 1
+                plannedThisRound = []
+                # Which modifier touched which element in THIS round. Two modifiers mutating one
+                # element within a round is a conflict even when their declared domains are
+                # disjoint -- e.g. one deleting what the other just created -- and the result
+                # depends on their order, which is exactly the kind of thing that must not decide a
+                # simulation quietly.
+                touchedBy = {}
```
alexdummer: **Possible bug: `touchedBy` conflict guard is reinitialized every round, so cross-round conflicts on the same element are never caught.**

`touchedBy = {}` is reset at the top of each round of `updateTopology`'s loop, so it only detects two modifiers mutating the same element *within one round*. A modifier that mutates element X in round 1 and a different modifier that mutates the same element X in round 2 of the same call are never compared against each other.

Scenario: two custom model modifiers that don't override `declaredDomain()` (default: claims nothing, so the static `checkModelModifierDomains` never flags them) mutate the same element in different rounds of one `updateTopology()` call — one in round 1, the other reacting to it in round 2. Neither the static domain check nor the round-scoped `touchedBy` dict catches it, silently corrupting element-set membership across modifiers — exactly the failure class this pipeline was built to prevent.

edelweissfe/modelmodifiers/adaptivity/hadaptivity.py
```diff
@@ -381,11 +384,32 @@ def __init__(self, name: str, model: FEModel, journal: Journal, *args, **kwargs)
...
+        :meth:`~edelweissfe.modelmodifiers.base.modelmodifierbase.ModelModifierBase.plan`.
+
+        The decision is returned as octree eids rather than element numbers: eids are this
+        modifier's own stable identifiers, reproducible by its replay, whereas element numbers are
+        assigned by the model's allocator in an order that depends on what else minted.
+        """
+
+        # Nothing this modifier cares about changed since it last planned in this topology update
+        # -- another modifier's mutation. Returning None here is what lets the pipeline settle.
+        if change is not None and not (change.addedElements or change.removedElements):
```
alexdummer, line 412: **Cleanup/robustness: `plan()`'s early-out only checks whether *anything* changed anywhere, not whether the change touches this modifier's own domain.**

Elsewhere in this same PR (`surfacefacets.py:96`, `tie.py:424`) the equivalent early-out uses `touchesElementSet`/`touchesSurface` to check domain relevance. Here, `if change is not None and not (change.addedElements or change.removedElements): return None` only avoids re-triggering by accident of the separate `_lastRefinedTime` guard a few lines below, not by contract.

Scenario: before hAdaptivity's first successful refine in a given `updateTopology()` call, any unrelated mutation elsewhere in the model (e.g. the implicit `surfaceFacets` modifier retiling a contact surface) causes hAdaptivity to fall through to a full marker re-evaluation every round, since the `_lastRefinedTime` guard hasn't fired yet — fragile behavior that depends on guard ordering rather than the modifier implementing the 'don't react to unrelated domains' contract the pipeline is designed around.

tests/test_hadaptivity_restart.py
```diff
@@ -0,0 +1,197 @@
...
+#  This library is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU Lesser General Public
+#  License as published by the Free Software Foundation; either
+#  version 2.1 of the License, or (at your option) any later version.
+#
+#  The full text of the license can be found in the file LICENSE.md at
+#  the top level directory of EdelweissFE.
+#  ---------------------------------------------------------------------
+"""Round-trip test for the recorded topology history of an h-adaptive model (unlike
+tests/test_restart_integration.py's AMR case, which drives the same feature end-to-end through a
```
alexdummer: **Stale reference: docstring cites `tests/test_restart_integration.py` as existing end-to-end coverage, but that file isn't in this PR or the worktree.**

"unlike tests/test_restart_integration.py's AMR case, which drives the same feature end-to-end through a real solve" — no such file exists here (same claim also appears in `tests/test_blockamg_forcing_tolerance.py:30`).

Scenario: a maintainer reading this docstring believes end-to-end restart+AMR integration coverage already exists and passes, when in fact `writeRestart`/`readRestart` never serialize or replay `topologyHistory` at all (no caller of `replayTopologyHistory` outside unit tests). The referenced file belongs to a separate, unmerged commit chain, so the claimed safety net for the deliberately-deferred restart wiring is a dangling reference, not real coverage.

edelweissfe/models/femodel.py
```diff
@@ -82,19 +90,439 @@ def __init__(self, dimension: int):
...
+                                    "model modifiers {!r} and {!r} both changed element {:} in round "
+                                    "{:} of one topology update. Whichever ran second silently won; "
+                                    "make their domains disjoint, or have one react to the other's "
+                                    "change in a later round instead of the same one.".format(
+                                        previous, name, elNumber, roundNumber
+                                    )
+                                )
+                    self.recordTopologyChange(roundNumber, name, modifier, plan, modelChange)
+                    plannedThisRound.append(name)
+                    changed = True
```
alexdummer, line 397: **Possible bug: `changed`/`plannedThisRound` are set unconditionally whenever `apply()` is called, regardless of whether the returned `ModelChange` is actually non-trivial.**

`changed = True` (and `plannedThisRound.append(name)`) fires whenever `modelChange is not None`, even if `apply()`'s change amounts to a no-op for this round.

Scenario: a modifier's `plan()` correctly returns non-None each round per the domain-touch contract, but `apply()` sometimes decides there's nothing to actually do. `changed=True` is set anyway on every such round, forcing an unnecessary DOF/equation-system rebuild; if this repeats it can exhaust `maxTopologyRounds` and raise `TopologyError` attributed to a modifier that never actually mutated the mesh in those rounds.

edelweissfe/models/femodel.py
```diff
@@ -110,8 +538,6 @@ def notifyModelChanged(self, changeType, change: ModelChange = None):
             change = ModelChange(kind=changeType)
         change.version = self.topologyVersion
         self._changeLog.append(change)
-        for observer in list(self._modelChangeObservers):
-            observer.onModelChanged(self, changeType, change)
 
     def changesSince(self, version: int) -> ModelChange:
```
alexdummer, line 614: **Efficiency: `changesSince()` does a full unpruned linear scan of `_changeLog` on every call, and the log is never trimmed.**

Called once per modifier per round in `updateTopology` and once per registered `MeshDependent` per increment in `refreshIfMeshChanged`, with `_changeLog` only ever growing (append-only).

Scenario: in a long-running adaptive simulation with many increments and continuous remeshing, every round of every increment re-scans the entire history so far, making aggregate topology-update cost trend toward quadratic in total mutations instead of linear. Since entries are already version-sorted, a bisect on `version` would give O(log n + k) instead of O(n).

edelweissfe/models/femodel.py
```diff
@@ -82,19 +90,439 @@ def __init__(self, dimension: int):
...
+                    raise TopologyError(
+                        "model modifiers did not settle within {:} rounds; still planning in the "
+                        "last round: {:}. A modifier must return None from plan() once the change "
+                        "since its own last plan no longer touches its domain.".format(
+                            self.maxTopologyRounds, ", ".join(plannedThisRound)
+                        )
+                    )
+        return changed
+
+    def topologyFingerprint(self) -> str:
```
alexdummer, line 410: **Efficiency: `topologyFingerprint()` hashes the entire model on every single applied modifier decision, not once per full topology update.**

`recordTopologyChange()` calls `topologyFingerprint()` unconditionally after every applied plan; it sorts and hashes every element and node in the whole model. `verifyTopologyFingerprints` only gates the replay-time *comparison*, not this computation, so it always runs.

Scenario: on a large mesh (10^5-10^6 elements) undergoing incremental AMR that touches only a handful of elements per increment, each applied plan still pays a full O(N log N + N) sort-and-hash over the entire mesh — dwarfing the cost of the actual small edit when run over thousands of increments, contradicting the comment that this is 'cheap enough to leave enabled.'

edelweissfe/models/femodel.py
```diff
@@ -82,19 +90,439 @@ def __init__(self, dimension: int):
...
+
+        if not self._topologyOpen:
+            raise TopologyError(
+                "element {:} was deleted outside a topology change: only model modifiers may create "
+                "or delete elements, inside FEModel.topologyChanges()".format(elNumber)
+            )
+
+        del self.elements[elNumber]
+
+    def ensureSurfaceFacetModifier(self, journal: Journal):
```
alexdummer, line 277: **Altitude: the implicit `surfaceFacets` modifier's 'runs last' guarantee rests on Python dict-insertion order, with no explicit ordering concept on `ModelModifierBase`/`updateTopology` to enforce or even detect a violation.**

`ensureSurfaceFacetModifier` is appended at the end of `prepareYourself`, so it 'reacts to whatever other modifiers just did' only because it's inserted last into `self.modelModifiers` (an ordered dict) — nothing in the round loop encodes or checks priority.

Scenario: a future change that reorders or dynamically registers `modelModifiers` (alphabetizing, a config option, dynamic registration) silently breaks facet-refresh timing with no error — contact/tie facets could regenerate before the refinement they're meant to react to, undetected by any check in the pipeline.

alexdummer:
**Automated review finding (cross-file, not inline since the file isn't touched by this PR's diff):**

**`edelweissfe/generators/abqmodelconstructor.py` — manually-defined `*element` blocks bypass the new `createElement()` uniqueness guard.**

This PR adds a uniqueness guard in `createElement()`/`reserveElementNumbers()` (and a test, `test_creating_a_taken_number_raises`, presenting duplicate-number detection as an enforced invariant), but `abqmodelconstructor.py` (unchanged by this PR) writes manually-defined `*element` input blocks directly via `elements.update(currElDefs)` into `model.elements`, bypassing that guard entirely.

Scenario: an input file combines a `*modelGenerator` (numbering `1..N` via `reserveElementNumbers`, `executeAfterManualGeneration=False`) with a manual `*element` block whose labels overlap `1..N` — a common pattern for hand-added elements alongside a generated base mesh. The generator's elements are silently clobbered with no warning and no `TopologyError`, even though the PR's own test presents duplicate-number detection as an invariant that should now always be enforced.

*(Posted as a general comment rather than inline because this file has no diff footprint in this PR — GitHub's review API only accepts inline comments on lines within the PR's diff.)*

edelweissfe/modelmodifiers/adaptivity/hadaptivity.py
matthiasneuner: Confirmed and fixed in `0e132d84` -- and the failure mode turned out to be **worse** than described.

Reachability, probed empirically: the cascade mechanism is real (given a >=3-level face adjacency, `balance_2to1()` re-splits a cell it created earlier in the same call). But I could not reach that state through the live `plan`/`apply` path: `apply()` balances to convergence every call, and a graded mesh plus one level of marked refinement gives at most a 2-level jump -- one short of what a cascade needs. 150 randomized live-path histories (5 configs x 30 seeds, 6 rounds) plus a deterministic shrinking-box history produced 0 events. So **latent, not active**, consistent with an earlier reviewer's 7-case instrumentation.

The worse part: when the cascade is constructed directly, the branch's filter does not merely drop quietly. The common shape crashes with a bare `KeyError` in the faceMap loop; and when *every* child of a materialised cell is re-split -- so nothing downstream names the intermediates -- it silently drops, measured at **74 active octree cells vs 10 model elements, 64 gone, no exception**.

Fix: `_materialize` now builds the ancestor closure of every unmaterialised active leaf and processes it level by level, so a cascade of any depth flows through the same "split a materialised parent into its children" code. Per-level changesets are coalesced, which is what collapses an intermediate's create-then-remove into the direct parent->grandchild relation instead of leaving a phantom in both `addedElements` and `removedElements`. The single-level case is byte-identical (coalescing one change returns it unchanged).

Two loud backstops replace the silent filter: an active cell with no materialised ancestor raises, and `set(_eidToEl) != active` at the end of `_materialize` raises. `tests/test_hadaptivity_cascade.py` (4 tests) pins it -- I verified all 4 fail on the previous tip and pass with the fix.

edelweissfe/generators/microstructuregenerator.py
matthiasneuner: Confirmed and fixed. You are right that the PR fixed exactly this pattern for elements and left it standing for nodes -- verified there was no `reserveNodeNumbers`/`createNode` at all.

Added the node-side counterpart mirroring the element API, including topology-window enforcement: `reserveNodeNumbers` / `adoptSetupNodeNumbers` / `createNode`. Wired into `inputfilehelpers` and `cuboidlatticegenerator` alongside the existing element calls, and used at both `microstructuregenerator` sites.

Two related gaps found while doing it, deliberately left for their own change rather than widened into this PR: AMR's own node labels still come from the octree `NodeRegistry` (`_maxLabel + 1`), bypassing the new counter -- unreachable today since AMR seeds after all generators, but the two counters can drift; and `boxgen`/`planerectquad`/`pipegen`/`discreterigidbodygenerator` still mint node labels as `max(model.nodes)+1`. Both worth an issue.

tests/test_hadaptivity_restart.py
matthiasneuner: Confirmed and fixed -- this one is mine: I removed `tests/test_restart_integration.py` during the transplant (it drives `*output, type=restart`, which trunk lacks) and never grepped for references to it. The docstrings were promising a safety net that does not exist here, which is exactly as misleading as you say given that `replayTopologyHistory` has no caller outside unit tests.

Removed the claims from both `test_hadaptivity_restart.py` and `test_blockamg_forcing_tolerance.py`, and grepped the whole worktree for other references to nonexistent files -- none remain (`test_mpc_cached_condensation.py` is mentioned but explicitly as "(removed)", and that text is pre-existing on trunk).

edelweissfe/models/femodel.py
matthiasneuner: Confirmed and fixed -- `touchedBy = {}` was indeed inside the round loop. It now spans the whole `updateTopology()` call and stores `(name, round)`, so a cross-round collision is caught and the message names both modifiers, the element, and both rounds. The existing within-round test still passes; a new cross-round test covers the case you describe.

edelweissfe/models/femodel.py
matthiasneuner: Confirmed and fixed. An empty change is now treated as nothing having happened: no history record, no version bump, no `changed = True`, no `plannedThisRound` entry -- so it cannot force a rebuild or burn `maxTopologyRounds`.

"Empty" is defined as a property on `ModelChange` itself (`isEmpty`, over all nine collections; `kind` deliberately excluded) rather than inlined at the call site, so there is one definition to keep correct. This is not hypothetical: `surfacefacets.apply` returns an empty change when a plan names only stale recipes.

edelweissfe/models/femodel.py
matthiasneuner: Correcting the comment now; deferring the optimisation to the restart PR, where the replay path it serves actually lands.

You are right on both counts and the cost is measured: `topologyFingerprint()` runs unconditionally in `recordTopologyChange` -- `verifyTopologyFingerprints` gates only the replay-time comparison -- at **0.188 s on 64k elements / 69k nodes, per applied decision**. The "cheap enough to leave enabled" comment was simply false at that scale; both comment sites now state the measured cost and what the flag does and does not gate.

edelweissfe/models/femodel.py
matthiasneuner: Agreed, and deferred deliberately rather than silently. Measured 0.55 ms at 20k entries, so it is tolerable per call but aggregates toward quadratic exactly as you describe, and `_changeLog` is never trimmed. A bisect on `version` is the right fix (entries are already version-sorted). Filing it with the fingerprint cost against the restart PR rather than growing this one further -- say the word if you would rather have it here.

edelweissfe/modelmodifiers/adaptivity/hadaptivity.py
matthiasneuner: Fair, and I agree the asymmetry with `surfacefacets.py`/`tie.py` is real -- those use `touchesElementSet`/`touchesSurface` while `hadaptivity.plan()` only checks whether anything changed anywhere, so the guard ordering is load-bearing rather than the contract.

Not fixing it here, deliberately: making `hadaptivity.plan()` domain-aware changes when marker evaluation happens, which is a behaviour change to AMR in a PR whose job is to move the pipeline across unchanged. It deserves its own change with its own before/after on the AMR suite rather than riding along here.

edelweissfe/models/femodel.py
matthiasneuner: Agreed that the guarantee currently rests on insertion order with nothing detecting a violation. Not fixing it in this PR: an explicit ordering/priority concept on `ModelModifierBase` is a design addition, and the pipeline is already carrying a lot of new surface here.

Documented instead for now -- `topologypipeline.rst` states the ordering rule and why the reactive modifier must run last. Worth a follow-up issue with the domain-relevance point above; both are about making the pipeline's implicit contracts explicit.

matthiasneuner:
Completing the node-allocator adoption in `6dcb43ab`, rather than leaving it half-done.

The previous round added `FEModel.reserveNodeNumbers`/`createNode` for the two sites the review named, but four generators plus AMR's octree registry still minted labels themselves — the same asymmetry, one level down.

**Generators.** `boxgen`, `pipegen`, `planeRectQuad` and `discreteRigidBodyGenerator` now draw node labels from the allocator. The awkward case was `boxgen`/`pipegen`: they construct a `Node` for every grid position but only *register* a subset (a 20-node hex has no face or body-centre nodes), advancing the counter only for registered ones — so reserving the full grid up front would have shifted every label. Handled by counting the registered positions first and reserving exactly that batch. `discreteRigidBodyGenerator`'s explicit `start_label` still wins; only the fallback goes through the allocator.

**AMR.** `NodeRegistry` (`adaptivity/refinement.py`) minted `_maxLabel + 1` independently of the model counter — not reachable today, since AMR seeds after all generators, but the two could drift into a collision. It now takes an optional `reserve_labels` callable, which `hAdaptivity` supplies as `model.reserveNodeNumbers`; only the minting branch draws a number, so the common "coordinate already known" path consumes nothing. Without a callable it behaves exactly as before, so the octree layer stays standalone and gains no `FEModel` import.

**Node labels are unchanged**, which matters because `U.ref` comparisons depend on them. Verified independently of the change itself: dumped every node label, its coordinates and full element connectivity from trunk and from the branch, and diffed — **227 node records and 50 element records across 5 configurations, byte-identical**, covering C3D8, C3D20 (the partial-registration case), CPE4, CPE8, and two generators writing into one model (where the second must continue the first's numbering).

**Suites**, by name *and* status at `OMP_NUM_THREADS=1`: `marmot` 74P/2F/40S and `edelweiss-only` 44P/1F/3S, identical sets to trunk — no new failures and no new skips. `pytest tests/` 309 passed plus the 5 known pre-existing (#93). Pre-commit clean.

Two adjacent things found and deliberately **not** changed: `cuboidLatticeGenerator` still replaces `model.nodes` wholesale and re-labels (it does call `adoptSetupNodeNumbers`, so the allocator stays consistent), and `planeRectQuad` raises `NameError` for element types with neither 4 nor 8 nodes — pre-existing on trunk, unrelated to this PR.

PR #98, opened 2026-08-28
feat(numerics): direct-to-CSR assembly, opt-in AMG hierarchy sparsification

matthiasneuner:
## Summary
Two independent, additive pieces of numerics infrastructure -- neither is wired into any solver's default path yet:

**`DirectCSRAssembler`** (`edelweissfe/numerics/_csrcore.h`, `csrgeneratorv2.pyx`): scatters entity blocks straight into a CSR data array, bypassing the VIJ staging array a solver currently builds before every gather. At scale that staging array is the dominant single allocation (12+ GiB at 43,350 DOF) -- bypassing it is the entire point. It borrows an existing `CSRGenerator`'s pattern rather than deriving its own, so there is exactly one definition of the CSR pattern shared by both assembly paths, and it exposes:
- `assembleFromVIJ`: a one-shot COO-equivalence path, for validating the addressing against `CSRGenerator`.
- `beginAssembly`/`scatterBlock`/`reduce`: the fused path a threaded entity loop calls directly, with a `corePointer` property so a downstream package's own threaded kernel (EdelweissMeshfree's particle loop) can call `scatterBlock` without going through Python.

`CSRGenerator` gained `patternOnly=True` (skips the gather map entirely -- halves memory and lifts the 32-bit COO-pair-count cap the gather map's own index arrays impose) and `releaseGatherMap()` (give the map back once a `DirectCSRAssembler` has borrowed the pattern).

**`blockamg` AMG hierarchy sparsification/lumping** (`hierarchyDropTol`, `hierarchyDropLumping`): build each field's AMG hierarchy from a sparsified copy of its diagonal block, dropping numerically negligible off-diagonal entries before the hierarchy build. Only the preconditioner is affected -- the operator the Krylov method applies is untouched, so this cannot change the converged solution, only the iteration count. Measured up to **1.81x on the linear solve** on operators with many near-zero entries (meshfree RKPM discretizations are the motivating case). Both off by default. Also skips the off-diagonal split step entirely when there's only one field (pure no-op removal in that case).

## Provenance and scope
This was bundled with the (unrelated) `fix/timestepper-zero-increment-honesty` bugfix on one very old branch; see [#90](https://github.com/Edelweiss-Numerics/EdelweissFE/pull/90), now cleaned up to just that fix. This is the other half, extracted and rebuilt directly on top of current `next_v26.11` (which already includes `#95`'s topology pipeline -- this branch's own real fork point, once its dead ancestry is stripped).

Neither piece is consumed by any solver yet. `DirectCSRAssembler`'s intended first consumer is EdelweissMeshfree's particle assembly, which needs exactly this API (`get_include()`, `corePointer`, `scatterBlock`) and was previously stripped out of a Meshfree PR specifically because this EdelweissFE-side dependency didn't exist on trunk. Landing this closes that gap; wiring it up on the Meshfree side is separate follow-up work.

## Test coverage
**The source branch shipped this with zero dedicated tests.** `DirectCSRAssembler`'s own docstring names `assembleFromVIJ` "the equivalence path ... to validate the addressing ... on real models rather than argued about," but no such test existed anywhere in the codebase. Added `tests/test_direct_csr_assembler.py`:
- `assembleFromVIJ` vs. the staged `CSRGenerator`, on a synthetic COO system with a duplicate-entry CSR cell (two "entities" both contributing to the same DOF pair) -- the case that actually exercises whether the addressing agrees, not just whether it runs.
- The fused `registerEntities`/`beginAssembly`/`scatterBlock`/`reduce` path against the same reference, using the header's own documented column-major per-entity block convention.
- `releaseGatherMap` and `patternOnly`.

Also manually verified (not committed as a test, just spot-checked on xeon) that `assembleFromVIJ` agrees with the staged reference at `numThreads` = 1, 2, and 4.

## Test plan
- [x] `pre-commit` clean on every touched/added file
- [x] Builds cleanly on xeon (`marmotelement`, `csrgeneratorv2`, `pardiso`, `amgcl`; only the pre-existing optional `panuapardiso`/`klu` gaps, missing system headers, tolerated by this project's `optional_build_ext`)
- [x] `python -m pytest` (full suite, x9 and xeon): 313 passed, 5 pre-existing failures reproduced identically on unmodified `next_v26.11`, unrelated (`tests/`-not-in-CI drift, same class as [#93](https://github.com/Edelweiss-Numerics/EdelweissFE/issues/93))
- [x] `run_tests_edelweissfe ./testfiles/edelweiss-only/` (xeon): 2 pre-existing failures (`MeshPlot`/latex, `NEDParallel`/`OMP_NUM_THREADS>1`), 0 new
- [x] `run_tests_edelweissfe ./testfiles/marmot/` (xeon, 114 ran): 2 pre-existing latex failures, 0 new

### DIFF
--- a/doc/source/documentation/assembly.rst
+++ b/doc/source/documentation/assembly.rst
@@ -52,56 +52,29 @@ value array is ever materialised.
       - 4 + 8 bytes / nnz
       - 4 + 8 bytes / nnz
 
-The two entries in bold on the left are why this matters. At 43,350 DOF of a meshfree RKPM
-discretisation -- 1.72e9 COO pairs against 71.2 M nnz, i.e. 24 duplicate contributions per matrix entry
--- the value array is 12.84 GiB and ``gather_sources`` a further 6.42 GiB, against a 0.53 GiB result.
+The two entries in bold on the left are why this matters: the more duplicate contributions land on
+the same matrix entry -- which grows with the number of entities touching a DOF, not with the DOF
+count itself -- the larger the value array and ``gather_sources`` become relative to the CSR result
+they produce. Their size is set by the number of COO pairs; the CSR pattern's is set by ``nnz``. A
+discretisation where many entities contribute to the same entries (a wide stencil, or overlapping
+supports) can have the staging array and gather map dwarf the actual result by a wide margin.
 
 Why the direct path is also faster
 ----------------------------------
 
-Three effects, all measured at 43,350 DOF on 16 threads:
+Three effects:
 
-- **Nothing has to be cleared.** The staging array must be zeroed every Newton iteration; at this size
-  that is 1.7 s of writing zeros. The direct path clears only the private CSR copies.
-- **Nothing has to be re-read.** The gather random-accesses a multi-gigabyte array; the scatter touches
-  only the CSR data, a ~23x smaller working set.
-- **The evaluation itself is slightly faster**, because each entity writes into a small reused scratch
-  block that stays in cache instead of a slab of a multi-gigabyte array.
+- **Nothing has to be cleared.** The staging array must be zeroed every Newton iteration; that cost
+  scales with the number of COO pairs, not with ``nnz``. The direct path clears only the private CSR
+  copies, sized by ``nnz``.
+- **Nothing has to be re-read.** The gather random-accesses the staging array, which can be many times
+  larger than the CSR data the scatter touches instead.
+- **The evaluation itself is not slower**, because each entity writes into a small reused scratch block
+  that stays in cache instead of a slab of a much larger array.
 
-Measured assembly, both paths timed in the same process on the same increment:
-
-.. list-table::
-    :width: 100%
-    :widths: 40 20 20 20
-    :header-rows: 1
-
-    * - Component, per Newton iteration
-      - Stage-then-gather
-      - Direct
-      - Ratio
-    * - clear the staging array
-      - 1.853 s
-      - 0.157 s
-      - 11.8x
-    * - entity evaluation
-      - 3.199 s
-      - 3.046 s
-      - 1.05x
-    * - gather / reduce
-      - 1.688 s
-      - 0.099 s
-      - 17.1x
-    * - **total**
-      - **6.704 s**
-      - **3.301 s**
-      - **2.03x**
-
-.. note::
-
-   That 2.03x is the *assembly in isolation*. Quote it with care: repeated later on the same data it
-   came out at 1.75x, because the gather's cost over a multi-gigabyte array varies with the machine's
-   memory state between runs, and it is the largest term the direct path removes. The honest figure is a
-   range, 1.75-2.03x.
+The relative sizes of these three effects depend on how many duplicate contributions land per matrix
+entry for a given discretisation and thread count -- measure on the model at hand rather than assuming
+a fixed ratio.
 
 The one definition of the pattern
 ---------------------------------
@@ -113,9 +86,8 @@ cannot drift apart. The generator must outlive the assembler.
 
 The offset map is built by binary-searching each ``(row, col)`` pair in the borrowed pattern, and stores
 the *within-row* position, which is why 2 bytes suffice: row length is set by the stencil, not by the
-problem size. Measured on an RKPM discretisation the largest within-row offset was 1,727 against the
-65,535 a ``uint16`` allows, and the bound stays in that region even as the support radius grows. A pair
-that cannot be mapped, or an offset that does not fit, raises rather than being silently misplaced.
+problem size, so it stays well within a ``uint16`` even for a wide stencil. A pair that cannot be
+mapped, or an offset that does not fit, raises rather than being silently misplaced.
 
 Building only the pattern
 -------------------------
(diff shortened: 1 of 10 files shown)


edelweissfe/numerics/csrgeneratorv2.pyx
```diff
@@ -228,3 +297,136 @@ cdef class CSRGenerator:
...
+        return self.csrMatrix
+
+    def registerEntities(self, long[::1] mapStarts, int[::1] nDofs):
+        """Give each entity its slice of the offset map, once per connectivity change.
+
+        ``mapStarts[e]`` is entity e's offset into the VIJ ordering -- the same value the DofManager
+        already records in ``idcsOfHigherOrderEntitiesInVIJ`` -- and ``nDofs[e]`` its local DOF count.
+        """
+        cdef int[::1] I = self._I
+        self.asm_.registerEntities(&mapStarts[0], &nDofs[0], mapStarts.shape[0], &I[0])
```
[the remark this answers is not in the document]
matthiasneuner: Fixed in `d85ce389`: `registerEntities()` now raises `ValueError` if `mapStarts` and `nDofs` lengths differ, before either is handed to C++ as a raw pointer. Added a regression test (`test_registerEntities_rejects_mismatched_mapStarts_and_nDofs_lengths`).

edelweissfe/numerics/_csrcore.h
```diff
@@ -258,4 +334,320 @@ class CSRCore {
...
+    if ( !useAtomics ) {
+      // one copy per thread: each thread clears its own, which is both perfectly parallel and
+      // keeps the pages it will later write on its own NUMA node
+#pragma omp parallel num_threads( nThreads )
+      {
+        double* my = priv[omp_get_thread_num()].data();
+        std::fill( my, my + nnz, 0.0 );
+      }
+             }
+             else {
```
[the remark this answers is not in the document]
matthiasneuner: Fixed in `d85ce389`. Turned out to be more interesting than a stray formatting slip: this project's pinned clang-format (v16, via pre-commit) actually *reintroduces* that exact mis-indentation on every run -- it's a known v16 limitation with a `#pragma omp` line immediately preceding the closing brace of an if-statement's compound body (confirmed by running the pinned hook against a manually-corrected version and watching it revert). Since a hand fix wouldn't survive the next `pre-commit`/CI pass, wrapped just that if/else in `// clang-format off` / `// clang-format on` with a comment explaining why, rather than leaving it to get re-mangled.

doc/source/documentation/assembly.rst
```diff
@@ -0,0 +1,218 @@
...
+and EdelweissFE provides both. Which one is appropriate is a memory question far more than a speed one,
+so this page describes what each costs as well as what each does.
+
+Both paths are in ``edelweissfe.numerics.csrgeneratorv2``, backed by the C++ header
+``edelweissfe/numerics/_csrcore.h``.
+
+The two paths
+-------------
+
+**Stage, then gather** (:class:`~edelweissfe.numerics.csrgeneratorv2.CSRGenerator`). Every entity writes
```
matthiasneuner, line 14: I think we still have the old v1 ?

doc/source/documentation/assembly.rst
```diff
@@ -0,0 +1,218 @@
...
+      - --
+      - **2 bytes / pair**
+    * - private CSR copies
+      - --
+      - ``nBuffers`` x 8 bytes / nnz
+    * - CSR pattern and data
+      - 4 + 8 bytes / nnz
+      - 4 + 8 bytes / nnz
+
+The two entries in bold on the left are why this matters. At 43,350 DOF of a meshfree RKPM
```
matthiasneuner: Why is RKPM mentioned here? Also, the numbers are not really relevant for the doc

doc/source/documentation/assembly.rst
matthiasneuner, line 14: Is the name "CSRGenerator" still specific enough? Also, do CSRGenerator and DirectCSRAssembler follow a common interface?

matthiasneuner: You're right -- `edelweissfe/numerics/csrgenerator.pyx` (single-threaded, binary-search mapping) is still there, and it's intentional: it's a deliberate teaching reference, not used by any solver (verified: zero consumers anywhere in EdelweissFE or EdelweissMeshfree). Fixed in `2dc4f18f` to actually say so -- both in `doc/source/documentation/utils.rst` (retitled its section "Legacy Generator (teaching reference)" with an explicit note it's not used by any solver) and in the class's own docstring, which previously just said "Very fast and convenient!" with nothing marking it as the non-default path.

doc/source/documentation/assembly.rst
matthiasneuner: Fixed in `2dc4f18f`. Fair point on both counts -- fixed throughout the page, not just here (the same RKPM/hard-number pattern repeated 5 more times elsewhere in this doc; generalized all of them). Kept the qualitative explanations (what scales with COO pair count vs. nnz, why atomics cost what they cost) since those are genuinely useful reference content, but dropped every specific benchmark figure and the EdelweissMeshfree-specific framing -- neither belongs in EdelweissFE's own docs, and the numbers would just go stale.

doc/source/documentation/assembly.rst
matthiasneuner: Fair question on both counts.

On the name: `CSRGenerator` predates this PR (it's the existing high-performance generator documented under `utils.rst`) -- `DirectCSRAssembler` is the new sibling. I don't think it needs renaming: it still describes exactly what the class does (generates a CSR matrix from COO input), and it's the class most callers reach for by default. `DirectCSRAssembler` earns its more specific name because it's the special-purpose alternative, not the default.

On a common interface: no, they don't share one today, and I don't think they should be forced to. `CSRGenerator`'s primary API is "give me an updated `csr_matrix` from a VIJ array" (`updateInPlace`/`updateCSR`) -- a single call per solve. `DirectCSRAssembler`'s primary intended usage is the fused `registerEntities`/`beginAssembly`/`scatterBlock`/`reduce` path, called once per *entity* from inside a threaded loop -- there's no single "value array" it primarily consumes. `assembleFromVIJ` (a VIJ-array-in, CSR-out method that could plausibly satisfy a shared interface) exists only as an equivalence path for validation/measurement, not as the class's main entry point, so unifying on that method would misrepresent how it's actually meant to be used. They do already share informal conventions (`csrMatrix`, `memoryBytes`) where the concepts genuinely overlap.

PR #104, opened 2026-09-02
perf(solvers): explicit dynamics performance optimizations, adaptive mesh refinement, and energy balance

matthiasneuner:
> [!NOTE]
> This PR builds on top of PR #101 (`feat(solvers): resolve Dirichlet/multi-point-constraint overlaps instead of rejecting them`). The first 4 commits are from #101; commits specific to explicit dynamics start after merge commit `a4d348f0`.

## Summary

This PR delivers several performance, adaptivity, and robustness improvements for the explicit dynamic solver (`NonlinearExplicitDynamic` / `NED`):

### 1. Performance Optimizations
- **Vectorized Element Gathering**: Gather element chunks via single fancy-indexing instead of Python-level item-by-item loops.
- **Fast Entity Lookup**: Return plain ndarray views from the entity lookup path.
- **Force-Only Constraint Path**: Bypass tangent stiffness assembly in constraints for explicit dynamics solvers that only require nodal force evaluation.
- **Spatial Indexing for Contact**: Find closest-facet candidates using spatial index bounding.

### 2. Adaptive Mesh Refinement (AMR) in Explicit Dynamics
- Enable both initial and mid-analysis hBcadaptivity during explicit dynamic simulations.
- Track and report smallest integrating element mass across mesh topology changes to safeguard the critical stable time increment.
- Tolerate quadrature-level mass variations across topology updates.

### 3. Energy Balance, Correctness & Diagnostics
- **Energy Balance Tracking**: Compute and report full energy balance (kinetic energy, internal energy, work) and flag non-physical energy creation.
- **Kinetic Energy Computation**: Corrected NED kinetic energy calculation to span second-order velocity fields only.
- **Explicit Step Cutbacks**: Explicitly refuse sub-stepping cutbacks (which are invalid in explicit dynamics) with clear error reporting.
- **Detailed Driver Failure Reporting**: Report underlying step failure root causes rather than generic simulation failures.
- **Checkpoint/Resume Integrity**: Prevent duplicate leapfrog startup half-steps on resumed explicit runs, and restore all node-field entries.


matthiasneuner:
Superseded by #109.

This branch is CONFLICTING because #101 was **squash-merged** as `91552970`: none of its five commit hashes are on `next_v26.11`, but this branch still carries all five plus three separate merge commits of `feat/mpc-dirichlet-reconciliation`. A rebase replays those against a target that already has their content, so #109 is a transplant of only the genuinely unique commits onto the current tip instead.

The split, as agreed:

1. Marmot MAteRialMOdelingToolbox/Marmot#81 — mass-lumping weight + critical time step. **Merged.**
2. #107 — restart correctness (`ef44397c`, `dfc17e59`, `d6d1d7ae`). **Merged.**
3. #109 — the explicit-dynamics feature and its performance work, 21 commits from here plus one fix and one style commit.

Also worth recording: `2d1b2dad` from this branch is **already upstream** — it was folded into #101 when that was prepared for merge, so it is in neither #107 nor #109.

Suggest closing this in favour of #109.

PR #113, opened 2026-09-04
feat(contact): integrated (Gauss-point-to-segment) surface contact for quadratic elements

### LINKED ISSUE #112: contact: hexa20 corner nodes lift off in node-to-surface contact — the consistent corner load is tensile, so no per-node weighting can fix it
## Symptom

Simulations using 20-node serendipity bricks (C3D20) with `nodeToDeformableSurfacePenalty` show
spikes of surface deformation in the contact regions. Observed in `examples/AnchorPryOut`.

## Root cause: the corner nodes lift off, and they must

The consistent nodal loads of a *uniform* pressure `p` on a quad8 face of area `A` are

```
corner:   -p*A/12     (TENSILE)
midside:  +p*A/3
```

verified by direct quadrature of the serendipity shape functions. A unilateral per-node penalty
spring can only ever push, so it cannot deliver the corner load. The discrete solution does not
resolve this by carrying a mis-distributed load — it resolves it by **opening the corner gaps**.

Measured on a new matched 2x2 hexa20 contact patch test (two blocks of identical 1x1 footprint,
nu = 0, so the exact state is homogeneous uniaxial stress with a flat interface and a total
interface force of 360):

|                          | interface force | corners in contact | max slave gap | interface flatness |
|--------------------------|-----------------|--------------------|---------------|--------------------|
| C3D8, for reference      | 359.892         | --                 | ~1.2e-5       | 1.4e-17            |
| C3D20                    | 353.787         | **0 of 9**         | 2.396e-3      | 1.4274e-3          |

All nine interface corner nodes are open, with gaps of 1.4e-3 to 2.4e-3 — two orders of magnitude
above the midside nodes' penalty compliance of ~1.2e-5. The 12 midside nodes carry the entire
load. The measured interface undulation *is* that liftoff (corner-minus-midside displacement
offset 1.081e-3 of the 1.4274e-3 total), and the 1.7% deficit in transmitted force is the spurious
interface compliance those open corners represent.

This is the classical quadratic-element limitation of node-based contact (the reason commercial
codes discourage node-to-surface contact on 20-node bricks). The existing note in
`doc/source/documentation/contacttheory.rst` anticipated it qualitatively; this quantifies it.

## What does NOT fix it

Reweighting the per-node tributary areas. The optimum is reachable and provable — over all
resultant-exact, non-negative weightings the mismatch against the consistent loads is
`w_corner + A/12`, monotone in the corner share, so a *vanishing* corner share is optimal (this is
the modified serendipity shape function set of Puso, Laursen & Solberg, CMAME 197, 2008:
`Ntilde_mid = N_mid + 1/2 (N_c1 + N_c2)`, `Ntilde_corner = 0`), and it cuts the mismatch from `A/8`
to `A/12` per node.

But it changes nothing here, because **the corner nodes are already open in the baseline**, so
their tributary area is moot. Measured: the interface force moves by 0.007% and the
corner-minus-midside offset is unchanged to 0.07%. A flatness difference does appear
(1.4274e-3 -> 1.2281e-3) but it is **not** a weighting effect: sweeping the retained corner share
continuously between the two weightings gives a non-monotone, two-valued response that tracks
which of the facets sharing a vertex-coincident slave's projection point wins the closest-point
tie-break. It is an artifact, not an improvement.

(Implemented anyway as a default-off `nodalWeights` generator option on the local branch
`feat/contact-serendipity-weights`, since the same modified basis is a prerequisite for the dual
multiplier space of quadratic mortar. It is documented there as delivering no measured benefit.)

## What does fix it: integrated contact with parent-face distribution

Gauss-point-to-segment (GPTS): quadrature points on the slave surface, a pointwise non-negative
pressure, and — the load-bearing part — distribution via the **parent face's own quadratic shape
functions** rather than the flat facet's linear ones. Then a corner's nodal force is
`sum_q w_q * p_q * N_corner(xi_q)`, which is negative wherever `N_corner` is, so the corner
receives a tensile force derived from a non-negative pressure field, with no unilateral condition
imposed at the corner node at all.

Note that quadrature alone is not sufficient: the flat sub-facets' linear shape functions are
non-negative, so distributing with those would leave the corners lifting off exactly as now.

This premise is already proven in code (local branch, `edelweissfe/utils/parentfacegeometry.py`
plus `canonicalParentFace` in the surface element generator): integrating a unit pressure over the
existing 6-facet midside tiling and distributing with the quad8 parent basis recovers
`-A/12` per corner and `+A/3` per midside to 1e-13, with the 3-point triangle rule exact
(the integrand is quadratic in the facet's barycentric coordinates). It also works under
`triangulation=corner`, since the parent face is a property of the element, not of the tiling.

## Proposed scope

A sibling constraint (`surfaceToDeformableSurfacePenalty`), leaving `nodeToDeformableSurfacePenalty`
in place for comparison:

- [x] Parent-face geometry: canonical orderings derived from the existing face tables, quad8/quad4/
      line3/line2 shape functions, triangle and line quadrature rules, and the proof test above.
- [ ] The constraint: normal penalty contact, `sliding=small`, implicit and explicit paths.
      In small sliding the gradient is `w = kron([N^s(xi_q), -N^m(xi_m)], nbar)` — constant in the
      DOFs, zero Hessian, symmetric tangent, i.e. structurally the same algebra as the present
      formulation with the slave `1` replaced by the slave face's shape-function vector.
      One dense square block per quadrature point over `[slave parent nodes, master parent nodes]`,
      which is simpler than the present `K_pp/K_ff/K_pf/K_fp` split.
- [ ] Acceptance criterion, binary: the matched hexa20 patch test must pass — flatness -> ~1e-16,
      force -> 360, corner nodal forces genuinely negative — and the C3D8 patch tests must keep
      passing under the new constraint.
- [ ] Friction and augmented Lagrange at quadrature points (deferred; not needed for the pry-out,
      which uses neither). Note the real difficulty is state, not kinematics: quadrature points have
      no identity to key friction history on across an AMR retiling, unlike nodes.

## Boundary of the fix

GPTS removes the liftoff but is **not** mortar and will not make non-matching interfaces exact:
pointwise enforcement at Gauss points over-constrains (inf-sup / patch-test failure, El-Abbasi &
Bathe 2001), and the integrand is discontinuous inside a slave facet wherever the assigned master
facet changes, so that quadrature error does not shrink with more Gauss points. Exactness needs
segment clipping and weak enforcement with a multiplier field — full mortar, e.g. Puso & Laursen
2004 and, for the quadratic dual basis, Popp, Wohlmuth & Wall (SISC 2012).

Worth recording that `tie` has the same non-matching limitation (its docstring scopes itself to
"compatible discretizations"), so a single clipping + weak-projection layer would serve both
constraints. That, rather than contact alone, is what would justify the mortar investment.

## Workaround available today

A linear (C3D8) contact layer in the bearing zone, tied to the hexa20 bulk with the existing `tie`.
Linear faces have no negative consistent loads to demand, so there is nothing to lift off, and the
C3D8 contact patch test is already exact.

matthiasneuner:
Fixes #112.

## The problem

A node-based penalty scheme cannot transmit a correct pressure across a serendipity (quad8/hexa20)
face, and **no choice of nodal weights can repair it**. The consistent nodal loads of a uniform
pressure `p` on a quad8 face of area `A` are `-pA/12` at the corners and `+pA/3` at the midsides —
the corner load is *tensile*, while a unilateral spring can only push. The discrete solution
resolves the contradiction by opening the corner gaps.

Measured on a new matched 2x2 hexa20 patch test (exact solution: homogeneous uniaxial, flat
interface, total force 360):

| | interface force | corners in contact | max slave gap | interface flatness |
|---|---|---|---|---|
| C3D8, for reference | 359.892 | — | -1.2e-5 | 1.4e-17 |
| hexa20, `nodeToDeformableSurfacePenalty` | 353.787 | **0 of 9** | +2.396e-3 | 1.4274e-3 |
| hexa20, `surfaceToDeformableSurfacePenalty` | **359.892** | **all** | **-1.1996e-5** | **3.19e-16** |

All nine corner nodes lift off in the node-based case, with gaps two orders of magnitude above the
midsides' penalty compliance. The measured surface undulation *is* that liftoff, and the 1.7% force
deficit is its spurious compliance. This is the surface-spike signature seen in the explicit anchor
pry-out runs.

## The fix

`surfaceToDeformableSurfacePenalty` imposes unilaterality on the pressure **field** rather than on
the nodes: the penalty pressure is evaluated at quadrature points over the slave facets, where
non-negativity is the physically correct statement, and distributed with each side's **parent
element face** shape functions, so

```
f_a = sum_q w_q J_q p(g_q) N_a(xi_q)
```

is negative wherever `N_a` is. A tensile nodal force arising from a non-negative pressure field,
with no unilateral condition ever imposed at a corner node.

**Quadrature alone would achieve nothing** — the flat sub-facets' own linear shape functions are
non-negative, so distributing with those leaves the corners lifting off exactly as before. The
parent-face basis is the load-bearing half.

Every figure above is what theory demands, not merely an improvement:
- resultant `359.89203239028313` vs the C3D8 reference's `359.89203239028274` — 13 digits;
- gap `= -p/penalty` at *every* quadrature point, i.e. uniform penetration and hence uniform
  pressure, nothing lifted off;
- minimum slave nodal normal force `-29.99100269952982` against the analytic `-4pA/12 =
  -29.99100269919026` for an interior corner shared by four faces.

## Scope, and what this is not

- **`sliding=small` only**, and that is not a compromise but the regime where the formulation is
  clean: the gap is linear in the DOFs, the gradient `w = kron([N^s, -N^m], nbar)` is constant, the
  geometric Hessian vanishes, and the tangent is symmetric. `sliding=finite` is **rejected** rather
  than approximated — it needs the exact gap gradient and Hessian against a *curved* quadratic
  surface, whose closest-point projection is itself an iterative sub-problem.
- **Penalty only.** Friction and augmented Lagrange are not implemented (the node-based constraint
  keeps both). Per-quadrature-point state is carried so they can be added without restructuring;
  the hard part there is state, not kinematics — quadrature points, unlike nodes, have no identity
  that survives an AMR retiling.
- **Not mortar.** Pointwise enforcement at quadrature points still over-constrains a non-matching
  interface, and the integrand is discontinuous inside a slave facet wherever the assigned master
  facet changes, so that quadrature error does not shrink under refinement of the rule. Exactness
  on non-matching meshes needs segment clipping and weak enforcement through a multiplier field.
  `tie` shares that limitation, so one clipping layer would serve both — that, not contact alone,
  is what would justify the investment.
- The node-based constraint is **untouched**. On *matched* linear faces the parent-face basis
  coincides with the facet basis and the two agree to 13 digits — but see the correction below:
  that is not true on non-matching meshes.

## A negative result, kept deliberately

`nodalWeights=serendipityOptimal` (first commit) is the provably optimal *non-negative per-node*
weighting — over all resultant-exact non-negative weightings the mismatch is `w_corner + A/12`,
monotone in the corner share, so a vanishing corner share is optimal (the modified shape functions
of Puso, Laursen & Solberg, CMAME 197, 2008). **It delivers no measured benefit**, because a
quadratic face under near-uniform pressure opens its corner gaps regardless, which makes the corner
weight moot. It ships default-off and is documented as unproven. It is kept because the same
modified basis is a prerequisite for the dual multiplier space of quadratic mortar (Popp, Wohlmuth
& Wall, SISC 2012) — it is on that path, not off it.

Note for anyone reading the flatness column of its own patch test: the difference there is a
closest-point tie-break artifact, not a weighting effect. Sweeping the retained corner share
continuously gives a non-monotone, two-valued response. Recorded in the test header.

## Non-matching meshes, and a correction

A figure of the matched patch tests would show nothing — all four element pairings pass to machine
precision. The informative model is the **non-matching** one (`doc/figures/contact_nonmatching_comparison.py`,
figure committed and shown in the theory docs): box on box, 3x3 against 4x4 elements, exact solution
a flat interface at uniform `S33 = -360`.

| pairing | node-based RF · max\|Δu_z\| | integrated RF · max\|Δu_z\| | gain |
|---|---|---|---|
| hexa20 / hexa20 | 350.3 · 1.35e-3 | **359.892** · 5.57e-6 | 242x |
| hexa20 slave / hexa8 master | 357.7 · 2.25e-3 | **359.892** · 2.16e-6 | 1042x |
| hexa8 slave / hexa20 master | 348.5 · 1.91e-3 | **359.890** · 3.11e-5 | 61x |
| hexa8 / hexa8 | 358.6 · 2.72e-3 | **359.892** · 1.14e-5 | 238x |

**This corrects a claim I made above and in an earlier revision of this description.** I wrote that
linear faces have "nothing to gain". That holds only on a *matched* interface. On a non-matching one
the hexa8/hexa8 pairing improves by 238x, because integrating the pressure over the slave facet
approximates the transferred load better than lumping it to the slave nodes — independently of
negative corner weights. Serendipity faces are where node-based contact fails *qualitatively*;
non-matching meshes are where it fails *quantitatively*, at any element order.

Also visible, and predicted: `hexa8 slave / hexa20 master` is the weakest integrated case (61x). The
quadrature points live on the slave facets, so a coarser-order slave under-samples the master's
basis — the reason to put the finer surface on the slave side.

Master/slave symmetry was checked separately on the matched test: all four permutations pass to
machine precision, including with the hexa20 as **master**, where all 21 of its interface nodes stay
flat to 2.4e-15. Both sides are distributed through their own parent-face basis, so the fix applies
to whichever side carries the quadratic element.

## Performance

The per-point Python loop cost **29 us per contact point** for work that is two 8x3 matmuls and a
48-entry outer product — essentially all interpreter and numpy dispatch overhead. On the explicit
anchor pry-out that put contact at roughly half the wall clock, assembled **serially** (`NEDParallel`
overrides only `computeElements`, so its threads never touch constraints). Batching the active set:

| at pry-out scale (1014 slave facets, 3042 points) | before | after |
|---|---|---|
| all points closed | 88.59 ms | **2.50 ms** (0.82 us/point) |
| all points open | — | 0.70 ms (0.23 us/point) |

On the real model this brings the integrated formulation to **1.22x** the node-based cost per
increment (0.05580 s vs 0.04583 s), i.e. no practical penalty.

Two findings worth keeping, both from profiling rather than reasoning:
- the batched arithmetic was never the cost — `_activeBlockSizes`, which the implicit protocol
  requests twice per evaluation, dominated everything else combined. Now cached.
- the local DOF vector is one contiguous, unshared block per active point (points share nodes only
  when the *solver* scatters into the global vector), so the gather and the scatter are plain
  reshapes: no index arrays, no `bincount`, no `np.add.at`.

## Verification

- **12 unit tests** on the constraint, including a **central-difference tangent check** — done
  deliberately, because the patch test is nearly linear and converges in two iterations even with a
  wrong tangent — plus symmetry, explicit/implicit force equality, resultant-invariance over 1/3/6
  point rules, the analytic corner and midside loads, and every validation refusal.
- **A batched-vs-per-point equivalence test.** The per-point loop is retained as the reference
  implementation and is used whenever parent-node counts are non-uniform. Measured agreement
  **2e-15** relative on gaps, forces and tangent; exactly zero for the linear law. Needed because
  every ordinary model now takes the batched path and the loop would otherwise be unreachable in
  tests.
- **14 tests** on the parent-face machinery, including the proof that quadrature against the quad8
  parent basis recovers `-A/12`/`+A/3` to 1e-13, under both triangulations, and Simpson for the 2D
  `line3` edge.
- **Four new regression cases**: the hexa20 acceptance test, its `facetConsistent` baseline, the
  C3D8 patch test under the new constraint (a drop-in where the node-based one was already exact),
  and an explicit-dynamics case.
- Full `edelweiss-only` suite green, **no reference drift on any pre-existing case**.
- Docs: a new `contacttheory.rst` section deriving the mismatch, its optimum, and the integrated
  formulation with its limits, plus the constraint registered in `constraints.rst`.

`NEDSurfaceContact`'s reference was regenerated once: over 2000 explicit increments the 1e-15
summation-order difference of the batched path grows to ~1e-3, because `g < 0` is non-smooth and a
near-zero-gap point can flip state one increment early. **A 1e-3 disagreement in that test is
therefore not by itself evidence of a defect**; the implicit patch tests have no such amplification
and reproduce bit-identically. Recorded in the test's own header.

## Field status

Running the explicit anchor pry-out with both contacts switched over. Damage initiates at
`U = 5.359782e-03` with `RF = -404.7657`, against the node-based run's `5.359782e-03` /
`-404.7661` — six significant figures. That is expected and is the point: the corner-liftoff defect
is self-equilibrated, so it shows up in the pressure *distribution*, not the resultant. The
evidence for the fix has to come from the interface fields, not the load-displacement curve.

## Unrelated blocker found on the way

The **pure-Python C3D20 element cannot run explicit dynamics at all**: row-sum lumping of its
consistent mass matrix yields *negative corner masses* — the same serendipity property as this
issue, in the mass matrix instead of the contact loads, wanting the same kind of fix there
(Duczek & Gravenkamp's 1/3-2/3 lumping weights). Verified pre-existing: unmodified `NEDContact`
with only `elType` changed to `C3D20` fails identically with the node-based contact. Marmot's own
quadratic elements lump their mass themselves and are unaffected. This is why the explicit
regression case here uses C3D8. It deserves its own issue.


edelweissfe/constraints/surfacetodeformablesurfacepenalty.py
```diff
@@ -0,0 +1,884 @@
...
+        # is carried only by the closed ones: without a searchDistance every point is assigned, so
+        # these two numbers are what tell an explicit run whether its contact cost is doing work.
+        # The closed count is from the last force evaluation, which is the previous increment.
+        nAssigned = sum(1 for facetIdx in newAssignment if facetIdx is not None)
+        self.journal.message(
+            f"{nAssigned} of {self.nPoints} contact points assigned, "
+            f"{int((self._gapCurrent < 0.0).sum())} closed at the last evaluation",
+            self.name,
+            level=2,
+        )
```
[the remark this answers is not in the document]
matthiasneuner: Good catch — fixed in 7b091637, though I took the fix one level deeper than suggested.

Masking the count would have fixed the reported symptom but left the underlying array wrong.
`_gapCurrent` also backs the public `getGaps()`, and *its* callers cannot mask unassigned entries
because they have no access to the assignment. So the array itself is now reset: a point that loses
its master facet reports 0, the neutral value the array is initialized with. The closed count then
becomes correct by construction, which is why it deliberately carries no mask of its own — noted in
the comment so it doesn't get "fixed" again later.

Two related things came out of it:

- `getGaps()` now returns a copy, matching `nodeToDeformableSurfacePenalty`. Handing out the live
  array let a caller corrupt state that an augmented-Lagrange update would later read.
- The connectivity message is emitted only when the assigned/closed pair changes. An explicit run
  updates connectivity thousands of times and was logging two identical lines every time.

Covered by a new test that shrinks the search radius until nothing can be assigned and asserts the
gaps are cleared, plus one asserting the copy semantics. Note the defect is inert without a
`searchDistance`, since every point is assigned in that case — which is why the pry-out run
currently in flight is unaffected.

For a follow-up, deliberately not touched here: `nodeToDeformableSurfacePenalty` has the same
stale-gap pattern, and there `_gapCurrent` is read back by the incremental Uzawa update
(`nodetodeformablesurfacepenalty.py:1015`), so the consequence is a wrong multiplier rather than a
wrong log line.

edelweissfe/generators/surfaceelementgenerator.py
```diff
@@ -209,9 +373,34 @@ def buildContactFacets(model: FEModel, surfaceName: str, prefix: str, triangulat
...
+    if nodalWeights not in ("facetConsistent", "serendipityOptimal"):
+        raise ValueError(
+            f"surfaceElementGenerator: nodalWeights '{nodalWeights}' is not supported. Use "
+            "'facetConsistent' or 'serendipityOptimal'."
+        )
+
     if surfaceName not in model.surfaces:
         raise ValueError(f"surfaceElementGenerator: surface '{surfaceName}' is not defined.")
 
+    if nodalWeights == "serendipityOptimal" and triangulation != "midside":
```
alexdummer, line 385: The `nodalWeights='serendipityOptimal'` validation here only rejects 3D higher-order faces with `triangulation != 'midside'`; it never checks 2D quad8/CPS8/CPE8 edges. A 2D quadratic boundary surface with `triangulation='midside'` and `nodalWeights='serendipityOptimal'` produces 2 Line2 facets, which match neither branch that calls `_applyModifiedSerendipityShares`. Validation passes, no error/warning fires, and the option silently behaves like `'facetConsistent'`.

edelweissfe/constraints/surfacetodeformablesurfacepenalty.py
```diff
@@ -0,0 +1,904 @@
...
+            "weights": weights,
+            "slaveRef": slaveRef,
+            "masterRef": masterRef,
+            "nSlaveNodes": nSlaveNodes,
+            "nMasterNodes": nMasterNodes,
+            "nSlaveDof": nSlaveNodes * nDim,
+            "blockDof": blockDof,
+        }
+
+    def _applyConstraintBatched(self, U_np: np.ndarray, PExt: np.ndarray, K) -> None:
```
alexdummer, line 684: `SurfaceToDeformableSurfacePenaltyConstraint` never reads `facet.weightTransform`/`nodalAreaShares`, so `nodalWeights='serendipityOptimal'` is a silent no-op on this constraint type. If a user builds a master surface with `nodalWeights='serendipityOptimal'` and pairs it with this (new) constraint instead of the old node-based one, behavior silently diverges from the sibling constraint with no warning.

edelweissfe/utils/parentfacegeometry.py
```diff
@@ -0,0 +1,201 @@
...
+
+    Returns
+    -------
+    tuple[numpy.ndarray, numpy.ndarray]
+        The barycentric coordinates, shape ``(nPoints, nFacetNodes)``, and the weights, which sum to
+        one -- so an integral over the facet is ``facetMeasure * sum(weights * f(points))``.
+    """
+
+    rules = _TRIANGLE_RULES if nFacetNodes == 3 else _LINE_RULES
+    if nPoints not in rules:
```
alexdummer, line 196: `facetQuadratureRule` validates `nPoints` only against the rule-table keys, not against the polynomial degree actually required by the active `parentFaceShapeFunctions` branch. A user setting `nQuadraturePoints=1` for a quad8/hexa20 parent face gets no error, and the contact traction is silently under-integrated.

edelweissfe/generators/surfaceelementgenerator.py
```diff
@@ -171,7 +276,64 @@ def _assignQuadConsistentShares(quadFacets: list):
...
+    Applied per facet: the midside table lists every corner triangle as ``(m_prev, c, m_next)``, so
+    local index 1 is the corner and its two adjacent midside nodes on the parent face are exactly
+    local indices 0 and 2. The redistribution therefore never leaves the facet, and stays correct
+    for distorted and curved faces, whose four corner triangles have unequal areas.
+
+    Side effect: the corner nodes of the affected faces end up with a zero tributary area, i.e.
+    they are no longer contact points at all. This is intended (it is what removes the corner
+    over-constraint), but it also means they carry no penetration guard.
+
+    IMPORTANT, measured: no benefit has been demonstrated for this weighting. Wherever a quadratic
```
alexdummer, line 310: `_MODIFIED_CORNER_TRIANGLE_TRANSFORM`'s own docstring admits 'no benefit has been demonstrated for this weighting' (0.007% force change), yet it remains ~180 lines of hand-derived special-case machinery serving only the old node-based constraint. Future maintainers may assume this is load-bearing and preserve/extend it (e.g. for a new element order) instead of recognizing it's superseded by the new GPTS parent-face quadrature, which this PR says fixes the same problem at the root.

edelweissfe/constraints/surfacetodeformablesurfacepenalty.py
```diff
@@ -0,0 +1,904 @@
...
+
+            if bestFacet is not None and (self.searchDistance is None or bestDistance <= self.searchDistance):
+                newAssignment[p] = bestFacet
+                facet = self.facetElements[bestFacet]
+                # The clamped barycentric weights locate the closest point in the flat facet; map
+                # that location into the parent face's parametric space and evaluate the parent
+                # face's shape functions there.
+                parametric = bestWeights @ facet.vertexParametricCoords
+                self._frozenMasterShapeFunctions[p] = parentFaceShapeFunctions(facet.parentFaceType, parametric)
+                normal, _ = facetNormalAndMeasure(facetCoords[bestFacet])
```
alexdummer: `updateConnectivity` calls `facetNormalAndMeasure` once per contact point instead of once per distinct assigned master facet. Since `nPoints = nSlaveFacets * nQuadraturePoints`, multiple quadrature points sharing a master facet redundantly recompute the identical normal/measure every connectivity update, on a code path this file's own comments flag as performance-sensitive.

edelweissfe/constraints/surfacetodeformablesurfacepenalty.py
```diff
@@ -0,0 +1,904 @@
...
+
+        closed = gaps < 0.0
+        if not closed.any():
+            return
+
+        g = gaps[closed]
+        penaltyTimesArea = self.penalty * b["weights"][closed]
+
+        if self.type == "linear":
+            f_n = penaltyTimesArea * g
```
alexdummer, line 711: The linear/quadratic penalty force-stiffness law is duplicated 3-4 times (this file's batched and scalar paths, plus `nodetodeformablesurfacepenalty.py`'s `applyConstraint` and Uzawa update) with no shared helper. A future correction to the law (sign convention, added regularization) could be applied to one copy and missed in the others, silently diverging batched vs. scalar (or old vs. new constraint) results.

edelweissfe/constraints/surfacetodeformablesurfacepenalty.py
```diff
@@ -0,0 +1,904 @@
...
+            stiffness = penaltyTimesArea
+        else:
+            f_n = -0.5 * penaltyTimesArea * g**2
+            stiffness = -penaltyTimesArea * g
+
+        # w = kron([Ns, -Nm], nBar), per point
+        c = np.concatenate((b["Ns"][closed], -b["Nm"][closed]), axis=1)
+        w = (c[:, :, None] * b["nBar"][closed][:, None, :]).reshape((len(g), -1))
```
alexdummer, line 719: `_applyConstraintBatched` rebuilds the gradient vector `w` from scratch every Newton iteration even though `Ns`/`Nm`/`nBar` are frozen for the whole increment under `sliding='small'`. Only the 'closed' mask changes per iteration; recomputing the full outer product every iteration instead of re-indexing a once-per-increment cache wastes O(nActive*blockDof) work per iteration across the whole run.

alexdummer (commented):
Automated review findings (Claude Code /code-review). 7 verified issues below.

edelweissfe/generators/surfaceelementgenerator.py
matthiasneuner: Correct, and fixed in 51caee4b.

Confirmed the mechanism exactly as you describe it: `quad8` *is* in `_MIDSIDE_FACE_TABLES`, but its
groups are 2-tuples, so it yields two `Line2` facets and reaches neither the
`len(faceFacets) == 2 and all Tria3` branch nor the `== 6` one that calls
`_applyModifiedSerendipityShares`. The existing check only fires on
`triangulation != 'midside'`, so `midside` + `serendipityOptimal` on a CPE8/CPS8 surface validated
and then did nothing at all.

It now refuses, with its own check and its own test
(`test_serendipityOptimal_rejects_2d_quadratic_edges`).

One thing that came out of writing the refusal message, and that I think improves on "not yet
implemented": in 2D there is nothing to implement. A quadratic edge's consistent nodal loads under
a uniform pressure are Simpson's `(L/6, 2L/3, L/6)` -- **all non-negative**. There is no tensile
corner load in 2D, so the mismatch is removable exactly there, and a corner-share reassignment (the
instrument that exists because 3D serendipity corners want a load a unilateral spring cannot
supply) is simply the wrong tool. The refusal says so rather than implying a gap, and it is
recorded in `contacttheory.rst`.

edelweissfe/constraints/surfacetodeformablesurfacepenalty.py
matthiasneuner: Agreed, and fixed in 51caee4b -- as a refusal, not a warning.

You are right that nothing here reads `weightTransform`/`nodalAreaShares`, and right that the
consequence is a silent divergence from the sibling constraint on identical input. What made me go
to a hard error rather than a log line is that there is no path by which this constraint *could*
honour a per-node weighting: it distributes the pressure with the parent face's own shape functions
evaluated at the quadrature points. So the option is not merely unimplemented here, it is
meaningless here -- and the mismatch it exists to minimise is the one this formulation removes
outright, which is the whole argument of the PR.

There is also precedent for refusing rather than warning on this exact option:
`nodetodeformablesurfacepenalty.py:_validateMasterWeightTransforms` already refuses a
`serendipityOptimal` surface under `sliding=finite`.

Checked at construction on both sides (a `serendipityOptimal` *slave* surface would be just as
silently ignored), naming the offending facet and pointing at `facetConsistent`. Covered by
`test_serendipity_weighted_surfaces_are_rejected`.

A side effect worth noting for your other comment on `_MODIFIED_CORNER_TRIANGLE_TRANSFORM`: with
this in place the weighting machinery is now unambiguously scoped to the node-based constraint,
enforced rather than merely documented.

edelweissfe/utils/parentfacegeometry.py
matthiasneuner: Accurate, and deliberately not fixed in this PR -- I would rather do it properly than bolt on a
guard.

`facetQuadratureRule` does only check the rule tables, so `nQuadraturePoints=1` on a quad8 parent
face under-integrates with no complaint. The default is 3 and the schema field documents *why* 3
("integrates the parent face's shape functions exactly over a facet, which is what makes the
consistent nodal loads ... come out exactly"), so this is a footgun for someone who overrides it
rather than a wrong default.

One detail worth recording, because it explains why the existing coverage does not catch it: the
`test_resultant_is_independent_of_the_quadrature_rule` test passes at 1, 3 and 6 points, and it
should -- the weights sum to one, so the *resultant* is exact at any point count. Under-integration
degrades the **distribution** of that resultant over the parent-face nodes, not its total. So the
right fix is a degree-aware rule table (each parent face type declares the degree it needs, and the
requested rule is checked against it) plus a distribution-level assertion, not a one-line minimum.
That is a follow-up, not a line in this diff.

edelweissfe/generators/surfaceelementgenerator.py
matthiasneuner: I want to push back on this one, while agreeing with the concern behind it.

The docstring already opens its measured section with "IMPORTANT, measured: no benefit has been
demonstrated for this weighting", and the PR description keeps it as a deliberate negative result.
The reason it stays is not sentiment: the same modified serendipity basis
(`Ntilde_mid = N_mid + 1/2 (N_c1 + N_c2)`, `Ntilde_corner = 0`) is the prerequisite for the dual
Lagrange multiplier space of quadratic mortar (Popp, Wohlmuth & Wall, SISC 2012). It is on the path
to the segment-to-segment work this PR argues for, not off it -- and mortar is what would finally
fix the non-matching case that GPTS only improves.

Your actual worry -- a future maintainer treating it as load-bearing and extending it to a new
element order -- I do take. Two things now guard against it. The docstring's "no benefit
demonstrated" paragraph, and, as of 51caee4b, an enforced boundary: `surfaceToDeformableSurfacePenalty`
now *refuses* a surface carrying a weight transform (your comment above), so the machinery cannot
be reached from the new constraint at all. Its scope is no longer a matter of documentation.

If you would still like an explicit "do not extend this for a new element order -- it is retained
for the dual mortar basis, not for contact weighting" line at the top of the transform, say so and
I will add it. I would rather not delete it.

edelweissfe/constraints/surfacetodeformablesurfacepenalty.py
matthiasneuner: Good catch, fixed in 51caee4b.

You are right that the call sat inside the per-point loop while the normal is a property of the
facet, and that `nPoints = nSlaveFacets * nQuadraturePoints` makes that a factor-of-`nQuadraturePoints`
overcount on a path the file's own comments flag as hot. Memoised per facet index for the duration
of one update.

The one thing to check before doing it is aliasing: the memo hands the *same* array object to every
point projecting onto that facet, where each previously got its own. I went through every use of
`_frozenNormals` first -- there are four, and none of them writes through it: two reads (`nBar[k] =
...` in `_prepareActiveArrays`, `nBar = ...` in the scalar path) and the two assignments in
`updateConnectivity` itself. The sharing is therefore safe, and the comment says so explicitly so
that a future in-place normalisation of a normal does not become an action-at-a-distance bug.

edelweissfe/constraints/surfacetodeformablesurfacepenalty.py
matthiasneuner: Valid, and a follow-up rather than part of this PR.

Worth splitting the four copies into two cases, because they are not equally exposed:

* the two **inside this file** (batched and scalar) are already guarded --
  `test_batched_and_per_point_paths_agree` compares gaps, forces and the tangent between them at
  2e-15 relative, exactly zero for the linear law. A one-sided edit to either is caught.
* the divergence that nothing guards is **across files**, new constraint vs
  `nodetodeformablesurfacepenalty.py`'s `applyConstraint` and its Uzawa update. That is the pair a
  shared helper would actually protect, and it is also the harder extraction: the Uzawa copy
  carries the AL multiplier and the node-based one carries friction, so the shared piece is the
  normal law only.

Filing it as a follow-up. I would rather not do a cross-constraint refactor in a diff this size --
it would touch the node-based constraint's converged reference behaviour, which is precisely what
this PR has been careful to leave alone.

edelweissfe/constraints/surfacetodeformablesurfacepenalty.py
matthiasneuner: True as written, and I am leaving it -- with the numbers, since they are the reason.

`Ns`/`Nm`/`nBar` are indeed frozen for the increment under `sliding='small'` and only the `closed`
mask moves, so `w` could be built once and re-indexed. Two things make it a poor trade right now:

* on the **explicit** path, which is where this constraint's cost actually lands, there is exactly
  one evaluation per increment. A per-increment cache saves nothing there by construction.
* on the **implicit** path the whole batched evaluation is 2.50 ms for 3042 points (down from
  88.59 ms per-point), of which the outer product is a fraction; and the cache is not free -- it
  holds `nActive * blockDof` doubles live across the increment, roughly 50 MB on the anchor pry-out
  at the shape this constraint's blocks take.

Profiling on this constraint has already produced one surprise in the opposite direction --
`_activeBlockSizes`, pure bookkeeping, dominated all the contact arithmetic put together -- so I
would want a measurement rather than a reasoned-about win before adding a cache with a memory cost.
Filing it as a follow-up with that note attached.

matthiasneuner:
## The red pipeline was a flake, and root-causing it corrected this PR's own explanation

Pushed as 52360ab5 (deck + coverage) and 51caee4b (the three review fixes). Flagging the first one
prominently because it changes a claim I made in the description above, and it changes the
`NEDSurfaceContact` deck that was reviewed.

**The symptom.** `NEDSurfaceContact` failed only in the with-Marmot job, 4 pass / 3 fail across the
branch -- including a pass to fail flip across a **docs-only** commit (eff4bd11 to 32872bd6, one
`.rst` file). `next_v26.11` green throughout. So: nondeterministic, not a regression.

**The explanation in the test header was wrong.** I wrote that the sensitivity was the batched
path's summation order amplified through the non-smooth `g < 0` activation. It is neither. The
switch is **master-facet reassignment** under `sliding=small`. Perturbing the penalty by 2e-16
relative (`5e4` to `50000.00000000001`) and running to completion:

| configuration | facet reassignments | final divergence |
|---|---|---|
| `nQuadraturePoints=3` (as reviewed) | 166 | **5.28e-4** |
| `nQuadraturePoints=1` | 0 | 6.11e-16 |
| `nQuadraturePoints=6` | 9 | 4.16e-16 |
| `contact-update-frequency=0` | -- | 3.82e-16 |
| node-based, `sliding=small` | 0 | 6.94e-16 |
| node-based, `sliding=finite` (as `NEDContact` actually runs) | n/a, no freeze | 4.16e-16 |

against a fixed 1e-6 absolute runner tolerance.

**My first control was not a control.** I had cited `NEDContact` as showing that the node-based
constraint on the same model does not amplify. It does not set `sliding`, and the node-based default
is `finite` -- no frozen projection at all. Re-run under `sliding=small` it *does* inject a jump at
every update once contact engages (up to 4.8e-6), and still reproduces to 6.9e-16, because its 16
slave nodes never cross a master element boundary where 54 quadrature points do. Re-projection is
not the chaotic ingredient; the discrete facet switch is.

**No defect in the constraint.** All 166 crossings are across genuinely different parent faces --
zero within the same parent face, which is the case that would have indicated a bug in the
flat-facet-barycentric to parent-face-parametric mapping. Median gap jump at a crossing is 1.2e-8.
The tail is the frozen facet normal changing across a *deformed* master element boundary
(|dn| up to 5.2e-2, |dgap| up to 4.8e-4, ~1.3% of the gap) -- the genuine kink between two
now-non-coplanar element faces, i.e. a property of facet-based contact, not of this implementation.

**What changed.** The deck pins its connectivity (`contact-update-frequency=0`) and its reference is
regenerated; the same perturbation now ends at 3.8e-16 in both directions and the case is
bit-identical at OMP=1/2/4/8. Pinning costs ~1% in the monitored results (`dispLowerMax` 0.03877 vs
0.03842) -- the small-sliding approximation error the option exists to trade away. Note that
picking a "lucky" quadrature rule would *not* have been a fix: 1 and 6 being stable is this mesh's
accident.

The reassignment path moved to two unit tests that are sharper than the deck ever was. On a flat
master a tangential slide cannot change the normal gap (`nbar` constant, `nbar . x` invariant under
tangential motion), so whatever the search returns must reproduce it exactly. Both assert
bit-identity -- within one parent face, and across two parent faces of a two-element master. Both
measure 0.0.

`contacttheory.rst` gains a subsection separating what throttling the connectivity search costs in
*reproducibility* from what it costs in *accuracy*; the existing text argued the accuracy case and
left the other half unsaid.

**Follow-ups deferred out of this PR**, all answered on their threads: the quadrature
degree/rule check, a shared normal penalty law across the two constraints, and caching `w` per
increment. Plus, from earlier: the facet-normal discontinuity across master element boundaries under
small sliding, whose standard remedy is averaged/smoothed nodal normals -- that one is a formulation
improvement, not a cleanup.

`edelweiss-only` suite green, no reference drift on any pre-existing case.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

edelweissfe/utils/parentfacegeometry.py
matthiasneuner: Filed as #116, with the resultant-vs-distribution trap and the degree-aware rule table recorded there.

edelweissfe/constraints/surfacetodeformablesurfacepenalty.py
matthiasneuner: Filed as #117, including the note that the two in-file copies are already guarded by the equivalence test and the cross-constraint pair is the one that needs the helper.

edelweissfe/constraints/surfacetodeformablesurfacepenalty.py
matthiasneuner: Filed as #118, with the measurements and the memory cost, flagged as a candidate to be measured rather than a known win.

matthiasneuner:
**Follow-ups filed**, so nothing above is left as a promise:

| | |
|---|---|
| #116 | `nQuadraturePoints` is not checked against the parent face's required degree |
| #117 | the normal penalty law is duplicated across both contact constraints |
| #118 | the gap gradient `w` could be cached per increment (measure first) |
| #119 | the frozen facet normal is discontinuous across master element boundaries under `sliding=small` |
| #120 | pure-Python C3D20 cannot run explicit dynamics -- row-sum lumping gives negative corner masses |

#119 is the only formulation-level one, and it is what the `NEDSurfaceContact` investigation
turned up: a limitation both constraints share, with the remedy (averaged nodal normals) and the
caveat that it interacts with the small-sliding argument for a symmetric tangent. #120 is the
pre-existing blocker noted in the description above, verified independent of this PR.

Pipeline is green on 4068f122, all four checks, and all eight review threads are answered.

PR #122, opened 2026-09-05
docs(examples): four-variant bonded anchor pry-out example (implicit/explicit x NTS/GPTS)

matthiasneuner:
A bonded M20 anchor at h_ef = 80 mm, loaded sideways through a fixture plate until the concrete
breaks out — set up **four times**, so that what changes between the variants is visible and what
does not is shared.

|  | node-to-surface contact | Gauss-point-to-segment contact |
|---|---|---|
| **implicit static** | `pryout_implicit_nts.inp` | `pryout_implicit_gpts.inp` |
| **explicit dynamic** | `pryout_explicit_nts.inp` | `pryout_explicit_gpts.inp` |

damage comparison

## Why this one

`examples/AnchorPryOut` already has the physics, at 280 k dof and a day per run — too big to open,
change something, and see what happens. This is the same model at **~29 600 dof and about half an
hour per variant**, which is the size at which people actually learn a code.

It exercises, together rather than one at a time: gradient-enhanced damage (`gcdp` on `GC3D20R`),
five tie constraints, two penalty contacts, and **live h-adaptivity** — the mesh refines *during* the
run wherever damage has started, and the ties and contacts are rebuilt around it each time.

## All four reach the same displacement

That is the point of the comparison, and it is why the explicit variants use ×100 mass scaling and a
0.024 s ramp: implicit against explicit at **equal load**, not at equal time.

| variant | wall time | refinement occasions | final concrete elements | final U |
|---|---|---|---|---|
| `implicit_nts` | ~26 min | 47 | 2201 | 1.500 mm |
| `implicit_gpts` | ~33 min | 51 | 2215 | 1.500 mm |
| `explicit_nts` | ~40 min | 12 | 2208 | 1.464 mm |
| `explicit_gpts` | ~30 min | 12 | 2215 | 1.464 mm |

All four arrive at essentially the same refined mesh from the same starting 1634 elements, by very
different routes. (16 threads, two variants at a time.)

## Redundancy

A variant deck is **six `*include` lines and a `*job`** — 28 lines including its header. Everything
shared sits in one place:

```
pryout_<solver>_<contact>.inp
├── materials_implicit.inc  or  materials_explicit.inc
├── model.inc                     mesh, sections, 14 interface surfaces, 5 ties
├── contact_nts.inc         or  contact_gpts.inc
├── adaptivity.inc                the live refinement marker
├── outputs.inc                   field outputs, Ensight, monitor
└── steps_implicit.inc      or  steps_explicit.inc
```

Changing contact formulation is one include line.

The boundary conditions are written twice, once per solver, rather than once — **`>>` option lines
cannot open an `*include`**: the parser raises `UnboundLocalError: options`, having lost the
enclosing keyword's context. A whole `*step` block inside an include works, which is what the step
files are. Worth fixing separately; noted here because it shapes the layout.

## The includes carry the reasoning, not just the values

The settings that are easy to get wrong are argued where they are set:

- **courant number 0.1, not the 0.8 default** — `l/c` is a linear-element formula and a hexa20's
  highest eigenfrequency is several times what it predicts. At 0.8 this model returns a reaction
  **26× too small and does not fail**: the run completes and the curve looks smooth.
- **Duvaut-Lions viscosity is zero under explicit integration** — at dT ≈ 3.4e-7 s nothing relaxes
  per increment, and a non-zero value makes the concrete respond nearly elastically: a plausible-
  looking run with no softening at all.
- **the AMR marker is damage, not stress** — deliberately the same in both solvers so the comparison
  is not confounded; a stress threshold ratchets towards refining everything under explicit dynamics.

## `plot_damage_comparison.py`

Renders the four damage fields on the symmetry plane with their load-displacement curves, and
reports **peak kinetic energy as a fraction of the external work** ∫RF·dU — which is what says
whether a mass-scaled explicit run may be read as a static result. The solver's own energy split
cannot answer it, since GCDP never populates a strain energy.

Here that ratio is **~2.2 %**, and the figure shows what that means: the damage fields agree closely
across all four, while the explicit reaction **rings around the implicit curves by roughly ±30 %**
once damage and contact are both active. The README says so rather than showing only the agreement,
and points at the trade-off — less mass scaling and a longer ramp buys a smooth history at a
proportional cost in increments.

One gotcha the script encodes: the Ensight export contains **one part per node set, 73 in all**, and
the concrete is *not* the largest by cell count. Selecting the largest block silently gives you a
part with no fields.

## The first commit

`feat(solvers)` makes the explicit mass-conservation tolerance an option, defaulting to today's
value. Without it the explicit variants **abort at their first refinement**: the bound is a hardcoded
1e-6 calibrated on the production mesh (4.63e-08 per refinement), and this 2× coarser mesh moves the
mass by 2.2e-06 — 48× more, which is how a quadrature truncation term behaves under coarsening, and
still five orders below the O(1) a real lumping error would give, which is the discriminator the
assertion's own message names.

## Notes

- The mesh is coarse **on purpose**: `l/h ≈ 0.12`, against 0.55 on the production mesh, which is
  itself not converged. The README says plainly that the contours are a demonstration of the
  machinery and not a prediction of the breakout cone.
- Independent of #113 and #121 — branched off `next_v26.11`, no shared files.


examples/AnchorPryOutCoarse/figures/damage_comparison.png
matthiasneuner, line 1: explicit shows some severe oscillations; much more than I think is accepteable

PR #129, opened 2026-09-06
docs(examples): split the full-scale pry-out into implicit and explicit variants

matthiasneuner:
`examples/AnchorPryOut/` held one monolithic implicit deck, `test.inp`. Running the same model
explicitly meant copying all 400 lines and editing four sections -- which is how the copies drifted:
**the header of every explicit revision claimed a quintic smoothstep amplitude while the block
underneath carried `f(t)='t**2'`, through three production runs.**

It is now laid out the way `examples/AnchorPryOutCoarse` is (#122). A variant is seven include lines:

```
pryout_implicit_gpts.inp ─┬─ materials_implicit.inc / materials_explicit.inc   differs
pryout_explicit_gpts.inp ─┼─ model.inc, contact_gpts.inc,
                          │  adaptivity.inc, outputs.inc                       SHARED
                          └─ steps_implicit.inc / steps_explicit.inc           differs
```

The mesh, the sections, all fourteen facet-surface generators, all five ties, the contact pairs and
the adaptivity criterion are the same files, not copies that have to be kept in step.

### The sharing was verified before it was factored, not after

The two monolithic decks were diffed section by section first. Six of ten sections -- MESH,
SECTIONS, CONTACT/TIE FACET SURFACES, TIE CONSTRAINTS, CONTACT and INITIAL MESH REFINEMENT -- were
**already code-identical**. Only materials, solver, field outputs and steps ever differed, and those
are exactly the includes a variant chooses.

### Verified against the decks it replaces

Both assembled variants reproduce the monolithic decks exactly:

| | monolithic | assembled |
|---|---|---|
| initial refinement | 110 marked, 11 176 -> 11 946, 574 hanging nodes | identical |
| contact points | 1188 + 1344 | identical |
| eliminated slave DOFs | 8804 | identical |
| implicit contact closure | 0 -> 160 -> 268 -> 505 | identical |
| explicit `dt_crit` | 1.725589e-07 | identical |

### Two deliberate choices

**No node-to-surface variants.** `AnchorPryOutCoarse` covers that comparison in half an hour. At ten
hours a run, shipping a configuration with the known hexa20 corner-liftoff defect invites spending a
day on it. `contact_gpts.inc` is factored out, so adding an NTS variant later is one file.

**The marker reads its own `alphaPForAMR` field output**, not the exported `alphaP`, following
`AnchorPryOutCoarse`'s convention that changing what the marker sees must not silently change what
is exported.

### `test.inp` is removed, its documentation is not

Nothing in the repo referenced the path. Its header carried the Abaqus translation provenance and
seven modelling decisions -- why the UEL became `GC3D20R`, what happened to the `COH3D8` cohesive
mortar layer, why node labels are offset by 100000, why Abaqus's `*Dynamic` step became a static one
here. Those move verbatim to `TRANSLATION_NOTES.md`, linked from the README with "read this before
changing the physics".

### The z-symmetry Dirichlet uses the mesh's own sets

An earlier revision of this mesh was short by 516 nodes on the symmetry plane and was patched by
including a generated `z_symm_all` set alongside. **That mesh has since been regenerated** (the
tracked version was last updated by #101). Checked against the shipped mesh: of 56 881 nodes, 5 259
lie on `z = 0`, **all 5 259 are members** of `z_symm_concrete`/`z_symm_steel`, none are declared that
are not, and the generated set is the same 5 259 labels. The patch is a duplicate and is not carried
here -- `zsymm_complete.txt` and `make_zsymm_complete.py` can be deleted wherever they still live.

### Note on the README's cost warning

The reference explicit run of this model took 32.2 h on 64 cores. The README says so up front and
points at `AnchorPryOutCoarse` for anyone who wants the same physics in half an hour.

Disjoint from #122 -- that adds `examples/AnchorPryOutCoarse/`, this changes
`examples/AnchorPryOut/`.


examples/AnchorPryOut/TRANSLATION_NOTES.md
matthiasneuner, line 1: Kill that file

PR #134, opened 2026-09-08
feat(solvers): damping and inertia for explicit dynamics -- the EdelweissFE side, plus a fracture-energy calibration

matthiasneuner:
Companion to **Marmot [#84](https://github.com/MAteRialMOdelingToolbox/Marmot/pull/84)** (artificial bulk viscosity and non-local micro-inertia). Must land together — the C++ signatures change, and adding a virtual to `MarmotElement` changes the vtable, so a stale `libMarmot` is not merely out of date, it is wrong. The extensions must be rebuilt against it.

This PR grew past what it opened as. It started as a one-line interface fix and now carries the EdelweissFE side of **both** explicit-dynamics devices in Marmot#84, because they are one effort and were calibrated together. What follows is the current scope, in full.

## 1. The interface fix this opened as

`MarmotElement::assignProperty(name, values)` took a bare `const double*` with no count, and `element.pyx` passed one. An element reading a fixed number of values from that pointer reads past the end of a short list **silently** — and the values come from a user-written `*elementProperty` block, which is exactly where a short list is likely. Marmot's side of the interface now carries the count; this passes it, and makes the array contiguous before taking the pointer.

## 2. Deriving the time integration instead of declaring it

An element reports two diagonals: the coefficient of each field's second time derivative (its inertia) and of its first (its damping). The solver reads both and needs nothing further — a field carrying an inertia gets central differences, one carrying only a damping gets forward Euler, and one update rule covers every second-order degree of freedom, reducing **bit-identically** to the undamped central difference wherever the damping is zero.

What the *diagnostics* need is a separate question, about units rather than physics: which entries of that inertia diagonal may be added to which. That is a property of the field, not of the analysis, so it is recorded once in `edelweissfe/config/phenomena.py` next to the field's tensorial order — a mass, a rotational inertia, or neither. Linear momentum sums the first kind; the energy balance the first two; conservation across a topology change is checked per field regardless.

> **Breaking input change.** `first-order-fields`, `second-order-fields`, `first-order-scheme` and `second-order-scheme` are **removed**, not deprecated: a declaration that agreed with the assembled operators was redundant, and one that disagreed was already refused. A deck carrying any of them now fails with `Invalid option`. **Deleting those lines is the whole of the migration** — nothing replaces them. Every deck in `testfiles/` is updated here.

## 3. A non-local field integrated as a damped wave equation

A first-order field's forward-Euler limit falls off with the **square** of the element size, so on a refined mesh it, not the mechanical field, bounds the increment. Giving it an inertia makes it second order — a damped wave equation whose limit falls off linearly — with the non-local viscosity keeping its value and changing role to the damping. Marmot supplies the inertia through the **material**: `getNonlocalMicroInertia`, read here from GCDP's property 21. It was the `nonlocal micro inertia` *element* property when this PR opened; Marmot#84 retired that name, because `m_k` is bounded by the non-local viscosity (`m_k ≤ η²/4`) and that bound is only checkable where both live in the same object. The two hyperbolic decks below move the value from `*elementProperty` into the material row; their `U.ref` files are untouched and still pass. That inertia is a time squared, so `0.5 m_k rate²` there is a volume rather than an energy: the field is registered as carrying a non-mechanical inertia, which keeps it out of both balances and onto a reported line of its own.

The damping is mass-proportional, so the update stays two lines of vector arithmetic.

## 4. Defects fixed along the way

* **`fix(adaptivity)`: refined children inherit their parent's named properties.** They inherited the section and nothing else, so a refined parent lost its bulk viscosity or micro-inertia — in exactly the region the refinement was requested for. Pre-existing in the bulk-viscosity feature, not introduced by the micro-inertia.
* **`fix(solvers)`: a run that has diverged to NaN fails the step.** Every ordering against a NaN is False, including the `Wext > 0.0` the energy guard is gated on, so a diverged run passed the check silently, kept integrating, wrote NaN into every output, and reported success.
* **`fix(elements)`: Marmot's warning channel is wired to stdout.** `MarmotJournal` holds an `std::ostream` on a **null streambuf** and nothing here ever called `setMSGOutputDirection`, so every warning Marmot has ever raised on this path was discarded before it could print.
* **`fix(elements)`: the lumped-assembly bindings had no exception handler.** `element.pxd` declared `computeLumpedInertia`, `computeLumpedDamping` and `computeCriticalTimeStepForExplicitDynamics` without `except +`. All three ask the material for the coefficients they assemble, and a material that refuses — too few properties, or (now) a micro-inertia above the `η²/4` its own viscosity admits — throws. Without a handler that C++ exception crosses into generated code that has none and reaches `std::terminate`, so a deck error aborted the process with no traceback instead of raising where the deck can be pointed at. The hazard predates the micro-inertia: `getNonlocalViscosity` has always thrown on a short property list.
* **Fixed in passing:** `secondOrderMomentum` summed fields of unequal dimension, which numpy *broadcasts* rather than rejects; and a micro-inertial `0.5 m_k rate²` was entering the energy balance.

## 5. Review round (2026-09-11)

* **A prescribed velocity stays prescribed under damping.** The damped update scaled Dirichlet-constrained degrees of freedom by `(1 - αΔt/2)/(1 + αΔt/2)` — with `P` zeroed there, nothing else is left in the expression — so a damped, prescribed DOF drifted off its increment every increment. Latent today only because `computeLumpedDamping` fills the non-local block alone, and live the moment any element reports mechanical damping. The velocities are restored after both updates.
* **A negative or non-finite lumped damping is refused where it is assembled.** The inertia check never saw it at a second-order DOF, where the damping enters only as the rate `C/M`: a negative rate amplifies the transient the damping exists to remove, and `αΔt/2 = -1` makes the update's denominator exactly zero.
* **The divisor is named by what it is** — an inertia at a second-order DOF, a damping at a first-order one — in the negative-value error and in the topology diagnostic, both of which still called it a mass.
* **Stale interface notes corrected** in the two parabolic GCDPNED decks.

## 6. Tests

* `testfiles/marmot/GCDPNEDExplicitHyperbolic` — the hyperbolic non-local field, agreeing with its parabolic sibling to 0.71 % in the displacement field.
* `testfiles/marmot/GCDPNEDExplicitHyperbolicAMR` — the hyperbolic field across a topology change. It still fails hard rather than drifting if a child loses its micro-inertia: the non-local field would then carry an inertia on some of its degrees of freedom and not others, which the solver refuses by name. Since Marmot#84 made the micro-inertia a material property the child inherits it through the **section**, by construction, so what this deck now covers is that guarantee rather than the `_elementPropertiesOf` bookkeeping — see *Still open*.

## Verification

Rebuilt against Marmot at `1665f7be` and run on one machine, `PYTHON_GIL=0`, `OMP_NUM_THREADS=4`:

* **Marmot: 50/50 ctests.**
* **`testfiles/marmot`: 2 failures**, `IndirectDisplacementControl` and `OutputManagers`, both `latex was not able to process the following string: b'lp'`. Environmental: reproduced outside EdelweissFE entirely, with a four-line matplotlib `usetex` snippet in the same environment. Every NED case passes — `GCDPNEDExplicit`, `GCDPNEDExplicitAMR`, `GCDPNEDExplicitHyperbolic`, `GCDPNEDExplicitHyperbolicAMR`, `NED`, `NEDLiveAMR`, `NEDLiveAMRStepEndsOffInterval`, `NEDParallel`.
* **`testfiles/edelweiss-only`: 2 failures**, `MeshPlot` (the same `usetex` breakage) and `NEDParallel` (the known OMP≥4 flake, which passes at OMP=1 — verified again here). `NED`, `NEDContact`, `NEDRestart1Write`, `NEDRestart2Resume`, `NEDSurfaceContact` and `TieNED` all pass.
* **The new test has teeth.** With the property propagation disabled, `GCDPNEDExplicitHyperbolicAMR` fails with the intended diagnostic: *"Field nonlocal damage was assembled an inertia on 20 of its 93 degrees of freedom and none on the rest."*

A model that declares no micro-inertia and no bulk viscosity takes none of the new paths and is bit-identical.

## Measured

The fracture-energy calibration behind these devices (20 runs, ~9 h, parabolic vs hyperbolic vs bulk viscosity across four mesh refinements) is written up separately and is **not** part of this PR — the example was dropped from the branch in `1a31f47a`. Its conclusions, for the record: parabolic and hyperbolic agree to five digits wherever both are stable; the hyperbolic scheme survives at `h = 0.625 mm` where the parabolic one diverges to NaN; bulk viscosity costs under 1 % of `G_f` at quasi-static loading; and the two devices do not interact.

## Still open

* **`hadaptivity._elementPropertiesOf` has lost its test.** It exists so a refined child does not lose its parent's named properties, and the micro-inertia was the only named property any deck assigned. Now that it rides along with the section, the remaining named property — `bulk viscosity` — is assigned by no deck, so that path has no coverage. Restoring it needs a deck with a *non-zero* bulk viscosity over an AMR element set; a zero one is indistinguishable from absent and would catch nothing. The fix in §4 stays: the mechanism is still live, just no longer exercised.

The **parabolic** non-local stability limit remains unchecked in code — mass scaling lifts the increment straight through it. Bounding it belongs in its own PR.


matthiasneuner:
Scope extended: this PR now carries the EdelweissFE side of both explicit-dynamics devices in Marmot#84, because they are one effort and are calibrated together.

**Added since the named-property count fix**

* **`fix(adaptivity)`: refined child elements now inherit their parent's named properties.** They inherited the section and nothing else, so a refined parent lost its *artificial bulk viscosity* -- in exactly the region the refinement was requested for. This is a pre-existing defect in the bulk-viscosity feature, not something the micro-inertia introduced.
* **`feat(solvers)`: a non-local field with a micro-inertia is integrated as a damped wave equation.** Declared in `micro-inertia-fields` alongside `second-order-fields`, because nothing in an assembled diagonal distinguishes a micro-inertia from a mass -- a density, a viscosity and a micro-inertia are all just positive numbers -- and the difference decides three diagnostics. Both directions of the declaration are checked against what the elements assembled. The damping is mass-proportional, so the update stays two lines of vector arithmetic.
* **`fix(solvers)`: a run that has diverged to NaN is now reported.** Every ordering against a NaN is False, including the `Wext > 0.0` the energy guard is gated on, so a diverged run passed the check silently, kept integrating, wrote NaN into every output, and reported success. Measured on the study below.
* **Fixed in passing:** `secondOrderMomentum` summed fields of unequal dimension, which numpy *broadcasts* rather than rejects -- a scalar field's fake momentum was being added to all three spatial components. And a micro-inertial `0.5 m_k rate^2` is reported on its own line instead of entering the energy balance, where a coefficient in seconds squared makes it a volume rather than an energy.
* **`examples/GCDP_implicit_vs_explicit_fracture_energy_calibration`** -- one bar, one loading history, five ways of integrating it: implicit, explicit with the non-local field parabolic or hyperbolic, and each of those with and without bulk viscosity. Pulled to the same 0.2 mm endpoint as `examples/AlphaP_AMR_Study`, so `G_f` is directly comparable to its 0.098 N/mm.
* **`testfiles/marmot/GCDPNEDExplicitHyperbolic`**, which agrees with its parabolic sibling to 0.71 % in the displacement field.

**Verification.** `edelweiss-only` 0 failures (the `NEDParallel` OMP>=4 flake is the known pre-existing one and passes at OMP=1); `testfiles/marmot` 82 passed, 0 failed; Marmot 50/50 ctests. A model that declares no micro-inertia takes none of the new paths and is bit-identical.

**Measured.** Parabolic and hyperbolic agree to 0.4 % in peak force and dissipated energy wherever both are stable, on an identical increment. From `h = 0.625 mm` the parabolic scheme diverges to NaN. Peak force converges onto the implicit reference: 67.21 N against 66.95 N. `G_f` did **not** converge in the earlier, shorter study (0.0515 -> 0.0816 against 0.098) -- whether that is the truncated softening tail or incomplete mesh convergence is what the shared 0.2 mm endpoint in the new study is there to settle, and it has not been run yet.

Requires Marmot#84. Adding a virtual to `MarmotElement` changes the vtable, so a stale `libMarmot` is not merely out of date, it is wrong -- the extensions must be rebuilt against it.

matthiasneuner:
### Update: the calibration study has now been run, and it needed fixing first

Two commits. Requires [Marmot#84](https://github.com/MAteRialMOdelingToolbox/Marmot/pull/84) at its current head.

**`fix(elements)`: wire Marmot's warning channel to stdout** — `MarmotJournal` holds an `std::ostream` built on a **null streambuf**, and nothing here ever called `setMSGOutputDirection`. So every warning Marmot has ever raised on this path was discarded before it could print, including the element's own long-standing `'sdv' is discouraged and deprecated`. Pointed at `std::cout`, where the solver's own output already goes. Both suites unchanged, and no new output — nothing in `testfiles/` triggers a Marmot warning.

**`docs(examples)`: calibrate the fracture energy at equal, quasi-static loading** — the study as committed could not measure a fracture energy, for three independent reasons, all found by running it.

*The crack was in the wrong place.* A gradient-damage bar localises at its own boundaries — the non-local field's zero-flux condition raises `eps_bar` there — and that beat the 0.7 % weakened central element in most runs. Which won varied with the mesh, so the implicit reference cracked at the clamp at two refinement levels and centrally at the other two, and its `G_f` jumped by a factor of two without converging: **0.1074 / 0.0493 / 0.0458 / 0.0876**. A boundary crack has only half its non-local band inside the bar. Fixed by strengthening one non-local length at each end, which removes the competition rather than out-shouting it — unlike a deeper notch it leaves the peak force a property of the material. The restraint is now quarter-symmetry instead of a fully fixed left face, which had been suppressing the Poisson contraction exactly where the reaction is measured; that alone took the implicit run from 2085 increments and a failure to 116 and a success.

*The loading was not quasi-static.* The explicit variants ramped over `1e-3 s` against the implicit `1 s`. The bar saw **1.2 wave transits** before the peak, the clamped-end reaction was still **0.50 N where statics required 27.56 N**, and the reported peak force was a transient — identical to three digits across runs whose material strength differed by 20 %. One duration for all of them is the only way the numbers mean the same thing.

*And the histories could not see their own peak.* `output-frequency` gates `fieldOutputController.finalizeIncrement`, so it is the sampling rate of the exported load-displacement history, not a logging setting. At its old value the explicit runs produced **ten-row** histories whose first sample already sat past the peak — reporting a softening-branch force as `F_max` and integrating `G_f` over ten points.

### Results

Twenty runs, ~9 h wall, every one reporting a kinetic share of **0.00 %**:

| nX | implicit | explicit-parabolic | explicit-hyperbolic | +bv (hyperbolic) |
|--:|--:|--:|--:|--:|
| 20 | 0.09969 | 0.09983 | 0.09983 | 0.10049 |
| 40 | 0.09842 | 0.09171 | 0.09171 | 0.09243 |
| 80 | 0.09115 | 0.08788 | 0.08788 | 0.08856 |
| 160 | 0.08759 | **DIVERGED** | 0.08604 | 0.08675 |

- **Parabolic and hyperbolic are identical**, not merely close — to all five digits at nX = 40 and 80.
- **The hyperbolic scheme still survives where the parabolic one does not**, at `h = 0.625 mm`, with and without bulk viscosity. Reported rather than passed off as success thanks to this PR's NaN guard (`4580ab38`).
- **Bulk viscosity now costs under 1 %** (+0.6 to +0.8 %). The 27–55 % measured before was a loading-rate artifact, and the two figures reconcile quantitatively.
- **The devices do not interact**: the bv increment is `+0.00072` with the micro-inertia and `+0.00072` without, at nX = 40.
- **All 18 stable runs crack centrally**, at the same site as the implicit reference.
- Peak force is a material property again: implicit 66.69–66.78 N, explicit 66.89–67.09 N, flat to 0.13 % across a factor of 8 in element size.
- `G_f` converges on both legs toward ~0.085–0.086, so `AlphaP_AMR_Study`'s **0.098 is a coarse-mesh value**, reproduced at nX = 20–40.

### Controls, because three parameters were chosen for run time

- **Raising `eta` tenfold to afford the increment changes nothing**: `G_f` 0.09171 against 0.09168 at nX = 40, 7.7 h of compute for the control alone.
- **The implicit step length is pseudo-time** — with decks differing in *nothing but* `stepLength`, `G_f = 0.098423` and `F_max = 66.7537` identically at 1 s and 1e-2 s. The implicit leg has no inertia and the material is rate-independent, so matching the explicit ramp to it matches a label, not a loading rate; what makes the comparison meaningful is `T/W_ext`.
- **The mass-scaling / `eta` levers are worth ~10×, and the limit is accuracy rather than stability.** At a kinetic share of 1.4 % — which sounds negligible — the peak force is 6–11 % high and non-monotone in the mesh. `mechanicallyDominatedViscosity()` records the rule the two knobs must satisfy together, and `--estimate` reports the increment count and whether the chosen `eta` clears its floor without running anything.

The write-up, figures and plotting script are deliberately **not** in this commit — they are going to a dedicated results repository.

### Still open

The **parabolic** non-local limit remains unchecked in code. This study showed it is a live hazard rather than a theoretical one: mass scaling lifts the increment straight through it, producing 960 divergence warnings, a NaN solution, and a run that still reports reaching the endpoint. Bounding it belongs in its own PR.

examples/GCDP_implicit_vs_explicit_fracture_energy_calibration/NOTES_hyperbolic.md
matthiasneuner, line 1: One file should be completely sufficient. Make it brief but exhaustive!

edelweissfe/solvers/nonlinearexplicitdynamic.py
```diff
@@ -36,6 +36,19 @@
 so a material that fails to integrate at the stable step is reported as an error rather than
 answered with a smaller step.
 
+**Micro-inertia.** A first-order field's forward-Euler limit falls off with the SQUARE of the
```
matthiasneuner: Why Micro-inertia? This seems oddly specific to certain physics

edelweissfe/solvers/nonlinearexplicitdynamic.py
```diff
@@ -290,6 +315,7 @@ class NED(NonlinearSolverBase):
     NEDOptions = {
         "first-order-fields": [],
         "second-order-fields": [],
+        "micro-inertia-fields": [],
```
matthiasneuner: Again. Why micro-inertia? Be general!

matthiasneuner:
Reduce the overly verbose docstrings/comments!

examples/GCDP_implicit_vs_explicit_fracture_energy_calibration/NOTES_hyperbolic.md
matthiasneuner: Done in ad8b313b: NOTES_hyperbolic.md deleted (it documented an earlier, superseded, narrower version of this study; its still-useful content -- the formulation pointer, the ringing-diagnostic tip -- is folded into RESULTS.md). RESULTS.md itself condensed from 444 to 190 lines, cutting the repeated restatement of the same numbers across the summary/results/controls sections while keeping the model, the parameter derivation, the results table, defects found, and reproduction steps.

edelweissfe/solvers/nonlinearexplicitdynamic.py
matthiasneuner: Agreed -- generalized in ad8b313b. The concept (a second-order field whose inertia is not a physical mass) isn't specific to non-local damage; only Marmot's current instance of it is. Renamed throughout: option micro-inertia-fields -> non-mechanical-inertia-fields, ids_microInertia -> ids_nonMechanicalInertia, _rawMicroInertia/_microDampingRate/_checkMicroInertiaAgainstDeclaration similarly -- following the naming nonMechanicalInertiaTotals() already used. Docstring section reworded to lead with the general concept and cite Marmot's micro-inertia as the concrete example, not the definition.

edelweissfe/solvers/nonlinearexplicitdynamic.py
matthiasneuner: Same fix as above (ad8b313b) -- this is the NEDOptions default dict entry for the option that got renamed.

matthiasneuner:
Addressed the four review comments (ad8b313b):

- Renamed the micro-inertia concept to **non-mechanical inertia** throughout the solver (option `micro-inertia-fields` -> `non-mechanical-inertia-fields`, `ids_microInertia`/`_rawMicroInertia`/`_microDampingRate`/`_checkMicroInertiaAgainstDeclaration` renamed to match, docstrings reworded) -- the capability isn't specific to non-local damage, only Marmot's current instance of it is. `nonMechanicalInertiaTotals()` already used this name; everything else now follows it. Updated the one test deck and example script that used the old option name.
- Folded `NOTES_hyperbolic.md` (a superseded, narrower version of this calibration study) into `RESULTS.md`, cutting it from 444 to 190 lines by removing the repeated restatement of the same numbers across the summary/results/controls sections -- one file, brief but exhaustive, as asked.
- On "reduce the overly verbose docstrings/comments": tightened the comments I touched today for the inertia/damping split (the `nonlinearexplicitdynamic.py` mass-assembly block) and the non-mechanical-inertia docstring section. I did not do a broader pass over the rest of the file's existing prose from earlier commits in this PR -- say if you'd like that too and which sections, since rewriting established reasoning at scale risks losing precision I can't independently verify.

Both suites still green: `testfiles/marmot` (GCDPNEDExplicit/-Hyperbolic/-AMR, NED, NEDLiveAMR*) pass, and Marmot's own ctest suite is unaffected by these changes.

edelweissfe/solvers/nonlinearexplicitdynamic.py
```diff
@@ -794,17 +832,45 @@ def solveIncrement(
...
+                #
+                # Damping is not optional for a non-local field made hyperbolic. Without it that
+                # field is a lossless wave equation driven by a slowly varying source: the
+                # transient minted at the start of the step, and again at every refinement, never
+                # decays, and the damage variable it drives accumulates on every overshoot rather
+                # than following the mean.
+                halfRateStep = 0.5 * self._dampingRate[self.ids_2nd] * dtAverage
+                V[self.ids_2nd] = (
+                    (1.0 - halfRateStep) * V[self.ids_2nd] + Minv[self.ids_2nd] * P[self.ids_2nd] * dtAverage
+                ) / (1.0 + halfRateStep)
```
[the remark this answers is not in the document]
matthiasneuner: Correct, and fixed in 773de852. Verified against next_v26.11: the undamped form there was `V += Minv * P * dt`, which at a constrained DOF added `Minv * 0 * dt` and left the prescribed velocity exactly — so this is new with the damping term, not pre-existing.

Restored rather than restricted: the prescribed velocities are saved in the Dirichlet loop and re-assigned after both kinematic updates, so one rule covers every scheme. That also picks up the first-order assignment `V[ids_1st] = Minv * P`, which loses a prescribed rate outright to `Minv * 0 = 0`. That one does predate this branch and no deck in the suite prescribes a first-order field, so the suite is unchanged by it.

It is latent today only because `computeLumpedDamping` fills the non-local block alone — but `GMCDPTriaxialTest` shows decks do prescribe `field=nonlocal damage`, and the comment right above the update says mechanical Rayleigh damping would need no code here, which is exactly when this would have broken every displacement BC. Worth having fixed before that day.

The two other line references (971, 1553) are the NaN guard and the damping-rate construction; neither touches a prescribed DOF. Only the site you flagged was affected.

edelweissfe/modelmodifiers/adaptivity/hadaptivity.py
```diff
@@ -656,6 +667,8 @@ def _materialize(self, model: FEModel, records: dict):
                     child = self._elementClass(self._elementType, elNumber)
                     child.setNodes([model.nodes[label] for label in e["conn"]])
                     self._sectionOf[parentEl].assignSectionPropertiesToElement(child)
+                    for elementProperty in self._elementPropertiesOf.get(parentEl, ()):
+                        child.assignProperty(elementProperty.propertyName, elementProperty.values)
```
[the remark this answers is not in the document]
matthiasneuner: Confirmed and added in e2697d27. `GCDPNEDExplicitHyperbolic` was indeed the only deck anywhere with an `*elementProperty`, and it has no topology change, so the inheritance path had no coverage at all.

`testfiles/marmot/GCDPNEDExplicitHyperbolicAMR` combines them: the hyperbolic bar with the micro-inertia property, refined once at one end before the increment loop. It fails hard rather than drifting if the propagation regresses — a child that loses the property leaves the non-local field carrying an inertia on the unrefined elements and none on the refined ones, and `_classifyFieldsByScheme` refuses a field whose own degrees of freedom disagree about which time derivative they carry, naming `*elementProperty` coverage as the usual cause.

edelweissfe/solvers/nonlinearexplicitdynamic.py
```diff
@@ -1351,10 +1451,67 @@ def buildEquationSystem(self, model: FEModel, step, previous: ExplicitSystem = N
             el.computeLumpedInertia(Me)
             M[el] += Me
 
+        # Each field's FIRST-derivative coefficient: zero mechanically, the non-local viscosity
+        # always (whether or not that field also has an inertia -- see computeLumpedInertia()
+        # above, the SECOND-derivative coefficient).
+        damping = self.theDofManager.constructDofVector()
+        damping[:] = 0.0
```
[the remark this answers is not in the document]
matthiasneuner: Fair — the description was accurate when the PR opened and then three scope extensions went into comments instead of into the body. Rewritten: it now states the full scope (the interface fix, deriving the integration scheme from the assembled operators, the hyperbolic non-local field, the four defects fixed along the way, this review round, and the tests), and calls out the breaking input change explicitly.

The removal of `first-order-fields` / `second-order-fields` / `first-order-scheme` / `second-order-scheme` is deliberate and stays a hard break rather than a deprecation: a declaration that agreed with the assembled operators was redundant and one that disagreed was already refused, so there is nothing for a compatibility path to preserve. The migration is to delete the lines, which is now stated both in the PR body and in the solver module docstring, where a user hitting `Invalid option` will look.

matthiasneuner:
### Review round addressed — 4 commits (`773de852`, `01f4436e`, `e2697d27`, `d280e240`)

Replied on the three inline threads. The collapsed *suppressed* comments were triaged too; these are the ones that were right, and what changed:

* **A prescribed velocity no longer drifts under damping.** Checked against `next_v26.11`: the undamped update added `Minv * 0 * dt` at a constrained DOF and left it exactly, so this is new with the damping term. The prescribed velocities are now restored after both kinematic updates — which also fixes the first-order assignment, where `V = Minv * P` loses a prescribed rate outright to zero. That one predates this branch, and no deck in the suite prescribes a first-order field, so the suite is unchanged by it.
* **A diverged run now fails the step** instead of reporting and continuing (`Wext`/`Wkin` non-finite → `StepFailed`). There is nothing left for the run to do, and an explicit scheme has no cutback to answer it with.
* **A negative or non-finite lumped damping is refused where it is assembled.** The `M < 0` check never saw it at a second-order DOF, where the damping enters only as the rate `C/M`: a negative rate amplifies, and `αΔt/2 = -1` makes the denominator exactly zero.
* **The divisor is named by what it is** in the negative-value error and the topology diagnostic — an inertia at a second-order DOF, a damping at a first-order one — and the two parabolic GCDPNED decks no longer claim `computeLumpedInertia` supplies the non-local viscosity.
* **The scope/acceptance criteria are rewritten into the PR body**, including the breaking removal of the four field/scheme options, which stays a hard break: a declaration agreeing with the assembled operators was redundant and one disagreeing was already refused. The migration — delete the lines — is stated in the body and in the solver module docstring, where a user hitting `Invalid option` will look.

**Not taken:** widening the NaN guard to `Wint` / `unaccounted` / the per-field non-mechanical rows. `unaccounted` is derived from `Wext` and `Wint`, and a state that makes `Wint` non-finite while leaving both integrated quantities finite is not reachable through the shared state; the two gates are the right ones.

**New coverage.** `testfiles/marmot/GCDPNEDExplicitHyperbolicAMR` is the only case anywhere combining a named element property with a topology change. It has teeth: with the propagation disabled it fails with *"Field nonlocal damage was assembled an inertia on 20 of its 93 degrees of freedom and none on the rest."*

**Verification** (rebuilt against Marmot `1665f7be`, `PYTHON_GIL=0`, `OMP_NUM_THREADS=4`): Marmot 50/50 ctests; every NED case passes in both suites, the new one included. The four remaining suite failures are not from this branch — `IndirectDisplacementControl`, `OutputManagers` and `MeshPlot` all die on `latex was not able to process the following string: b'lp'`, which reproduces outside EdelweissFE with a four-line matplotlib `usetex` snippet in the same environment, and `edelweiss-only/NEDParallel` is the known OMP≥4 flake, which passes at OMP=1.

PR #143, opened 2026-09-15
feat(modelmodifiers): snap AMR-refined boundary nodes onto an analytical cylinder

matthiasneuner:
## Summary

- Coarse HEX20 meshes represent a curved boundary (e.g. a borehole) as a polygon of flat facets. `Hex20Topology.subdivide()` evaluates the parent element's own isoparametric map when creating new nodes, so a flat parent facet only ever produces more flat sub-facets under AMR refinement — fidelity to the true curve never improves on its own.
- Adds `edelweissfe.modelmodifiers.geometry.surfacesnap`: a reactive `ModelModifierBase` that watches hAdaptivity's `ModelChange.faceMap` and projects newly-created boundary nodes on a tracked node set exactly onto an analytical cylinder (corners: unconditional radial projection; midsides: `straight` — the mean of their snapped corner endpoints — or `curved` — independently projected). Hanging-node slaves are detected via `HangingNodeConstraint.claimedSlaveNodes()` and excluded rather than snapped; a quality safeguard vetoes an entire round rather than let a snap invert or badly distort an element.
- Extends `ModelChange` with a `movedNodes` field and adds `ModelChangeType.GEOMETRY_CHANGE`, because a coordinate-only mutation was previously indistinguishable from an empty change and would never be recorded to `topologyHistory` — silently failing to replay after a restart. The modifier's own wall-face tracking lives in a real, modifier-private `NodeSet` (grown in `apply()` even on a round that snaps zero coordinates) rather than private Python state, so it replays correctly by the same mechanism hAdaptivity's own tracked node sets already use.
- Registered as `type=surfaceSnap` in `edelweissfe/config/registry.py`.

## Test plan

- [x] `tests/test_surfacesnap.py` (new, 7 tests) drives the real `*modelModifier` → `fillFEModelFromInputFile` → `model.updateTopology()` pipeline against a hand-built HEX20 sector mesh with a geometrically exact wall and deliberately flat everything else:
  - raw `Hex20Topology.subdivide()` alone leaves a real, measurable wall-node deviation from the true circle (sanity check of the fixture's premise)
  - `curved` mode lands every wall node (pre-existing and new) exact to `<1e-6`
  - `straight` vs `curved` mode produce the expected qualitatively different residual
  - `maxLevel=2` works cleanly
  - an asymmetric marker produces genuine hanging-node collisions at the wall's own rim, correctly skipped rather than corrupting the mesh
  - a full live run replays byte-identically onto a freshly-rebuilt model (`topologyFingerprint` equality), including the tracked wall set's own membership
  - a round that snaps zero node coordinates still produces a non-empty, replay-recorded change
- [x] Full `tests/` suite run before and after this work (with these changes fully `git stash`-ed out for comparison): 331 passed after vs. 329 before this work existed, 0 regressions. The only pre-existing failures (2 in `test_schemasurface.py`, 1 each in `test_inputlanguage_golden.py`/`test_stepoptions.py`) are golden-snapshot/grep-invariant drift from other in-flight work on `next_v26.11`, reproduced identically with this PR's changes removed.


matthiasneuner:
Addressed the Copilot review in d0011ec8. Went through each finding, verified against actual behaviour before changing anything:

**Fixed (real issues):**
- **hAdaptivity's stale mirror** (moderate): confirmed with a direct repro — `AdaptiveMesh` caches each element's coordinates at creation time and never re-reads `model.nodes`, so a later refinement of an already-snapped element would silently use pre-snap geometry. Added `HAdaptivityModelModifier.syncNodeCoordinates`, called from `surfaceSnap.apply()`. New test: `test_a_second_refinement_of_an_already_snapped_element_uses_corrected_geometry`.
- **Vetoed/collision nodes never retried** (moderate): real gap — a skipped node was permanently marked as confirmed wall membership with no path back to it, since the only trigger was `change.addedNodes`. Added a second tracked, modifier-private `NodeSet` (a "pending" set) that's retried on every later round that changes anything. New test: `test_a_resolved_hanging_collision_is_retried_and_snapped` (with its own restart-replay check).
- **`movedNodes`/`addedNodes` overlap in `ModelChange.mergedWith`** (nit): confirmed with a quick repro, real contract violation even though nothing currently consumes it. Fixed by stripping net added/removed labels from the merged `movedNodes`.
- **Missing validation on `radius`/`qualityDropThreshold`** and **`_wallSetName` collision with a user-declared node set**: both fair, both added.
- **Dead `PLAN_BOREHOLE_GEOMETRY_SNAP.md` references**: that's a local, intentionally-untracked planning doc (this repo's convention — same as several other `PLAN_*.md`/`HANDOFF_*.md` files never committed here), never meant to ship. Replaced with pointers to the actual Sphinx doc page.
- **`maxLevel=2` not really exercised** (nit): correct — the fixture's `initialOnly` marker can only ever reach level 1 in one `updateTopology()` call. Test now directly seeds a second refinement round (same technique `test_hadaptivity_cascade.py` uses) and asserts level 2 is actually reached.
- **Hanging-node collision test too weak** (nit): also correct, though the suggested "radial error" check turned out to be a false negative on this fixture (some already-near-exact nodes look fine even un-snapped). Fixed by asserting directly against the modifier's own recorded plan instead.

**Investigated, not changed:**
- **"Critical" empty-`coords` reshape bug**: does not reproduce. `np.array(()).reshape(-1, 3)` is well-defined (shape `(0, 3)`), and this exact path is already exercised by `test_a_round_that_snaps_nothing_still_replays`, which passes.

Full verification: `tests/test_surfacesnap.py` 9/9 (up from 7), full `tests/` suite 333 passed / 0 regressions.

PR #144, opened 2026-09-17
perf(solvers): evaluate each explicit element chunk in one Cython call

matthiasneuner:
## Summary
Profiling the pry-out explicit jobs found the parallel element loop wasn't scaling: every element did `scatter_P[element]`, a Python-level slice into one force buffer shared by all threads, whose reference-count bump under free-threading serializes across cores despite the loop looking parallel.

This gives each chunk its own thread-local force/gather buffers and moves the per-element loop into a single Cython call (`computeKernelsExplicitForChunk`) operating on raw pointers, with a per-element fallback for non-Marmot chunks or non-contiguous scatter slots. Chunking is also finer (16 chunks/thread instead of 4) since heavy return-mapping elements cluster spatially and unequal contiguous chunks were starving the load balance. Contact facets (`hasKernels=False`) are skipped in the explicit loop.

## Verification
- pytest: no new failures vs. a clean `next_v26.11` baseline (same pre-existing/flaky failures on both).
- Pry-out resumed-checkpoint run, 32 threads, 600 increments, against current upstream `next_v26.11`: elements 79.6 -> 44.8 ms, increment 95.3 -> 62.2 ms; RF/U bit-identical.


edelweissfe/elements/base/baseelement.py
```diff
@@ -310,6 +310,23 @@ def initialVelocity(self) -> np.ndarray:
         """
         return np.zeros(self.nDof)
 
+    @property
+    def hasKernels(self) -> bool:
+        """Whether this element contributes to the internal force and carries a state at all.
+
+        False for entities that live in the element container for their nodes and geometry
+        alone -- contact facets -- whose kernels, energy and state acceptance are no-ops. A
```
matthiasneuner, line 318: Stop mentioning specific numbers from random examples

edelweissfe/elements/marmotelement/element.pyx
```diff
@@ -442,3 +447,50 @@ cdef class MarmotElementWrapper:
...
+    cdef Py_ssize_t b
+    cdef double psi = 0.0
+    cdef double energy = 0.0
+
+    if offsets.shape[0] != n + 1:
+        raise ValueError("offsets must hold one entry per element plus the end")
+    if offsets[n] > Pe.shape[0] or offsets[n] > U.shape[0] or offsets[n] > dU.shape[0]:
+        raise ValueError("chunk buffers are shorter than the offsets require")
+
+    for i in range(n):
```
matthiasneuner, line 481: This is atrocious design! EdelweissFE is a teaching-oriented FE-code; This bullshit here is completely unreadable not not maintainable.

edelweissfe/solvers/base/parallelelementcomputation.py
```diff
@@ -39,6 +41,17 @@
 )
 from edelweissfe.timesteppers.timestep import TimeStep
 
+# The Marmot element extension is optional (a pure-Python build has none); without it every
+# chunk takes the per-element path.
+try:
+    from edelweissfe.elements.marmotelement.element import (
+        MarmotElementWrapper,
```
matthiasneuner, line 48: Horrible design!

edelweissfe/solvers/base/parallelelementcomputation.py
```diff
@@ -39,6 +41,17 @@
...
+        computeKernelsExplicitForChunk,
+    )
+except ImportError:  # pragma: no cover - depends on the build
+    MarmotElementWrapper = None
```
matthiasneuner, line 52: Absolutely no!

matthiasneuner:
Closing per review feedback — the Cython chunk kernel approach doesn't fit this codebase. Reworked as three smaller, plain-Python PRs (thread-local per-chunk buffer, dynamic chunk balancing, facet skip) with no Cython and no unconditional Marmot import. Replacements coming shortly.

PR #145, opened 2026-09-17
fix(numerics): restore pool-worker CPU affinity if something else pinned it

matthiasneuner:
## Summary
The element loop is a Python `ThreadPoolExecutor`, not an OpenMP team. If `OMP_PROC_BIND`/`OMP_PLACES` is set anywhere in the environment (a job script, a cluster module, a colleague's shell, a future BLAS/OpenMP library that pins on its own), `libgomp` pins the process's *initial* thread to one core the moment any extension linking it loads (`dirichlet`, `csrgeneratorv2`, `amgcl` in this stack) -- and every pool worker spawned afterward inherits that single-core mask. Nothing fails or warns; the run just gets ~6x slower in its biggest phase. Measured on live production explicit pry-out jobs.

This is belt-and-braces on top of removing the exports from our own job scripts: `edelweissfe/__init__.py` captures the start-up CPU affinity mask before any extension can load, and every pool worker restores itself to that mask on start, warning once via `RuntimeWarning` naming `OMP_PROC_BIND` as the likely cause. It cannot help if the OpenMP runtime is loaded before Python starts (e.g. via `LD_PRELOAD`) -- the captured mask would already be collapsed by then.

## Verification
- pytest: no new failures vs. a clean `next_v26.11` baseline.
- Emulation (libgomp loaded after package import under `OMP_PROC_BIND=close`): the guard's warning fires and the run holds unpinned speed (elements 80.5 ms, increment 97 ms) instead of collapsing; RF/U bit-identical either way.


edelweissfe/numerics/parallelizationutilities.py
```diff
@@ -29,9 +29,49 @@
...
+
+    A worker thread inherits the affinity mask of the thread that spawns it. With
+    ``OMP_PROC_BIND`` set, libgomp pins the initial thread to a single place as soon as it is
+    loaded, so every worker created after that runs on that one core and the free-threaded
+    element loop silently serialises -- measured on the anchor pry-out on LEO5: 21 of 32
+    workers confined to one physical core, the element phase 6x slower than unpinned. The
+    mask captured in :mod:`edelweissfe` at import predates any extension, so it is the
+    allocation the process was actually given (Slurm's cgroup included).
+    """
+    global _affinityWarned
```
matthiasneuner, line 52: Jesus, what the fuck>

matthiasneuner:
Closing per review feedback — the mask-capture/restore design is the wrong shape (silently overrides user-set affinity, doesn't cover every thread pool, no test). Replaced with a simple diagnostic warning in getThreadPool that tells the user what's wrong instead of trying to silently fix it. Replacement coming shortly.

PR #146, opened 2026-09-17
perf(solvers): keep the lumped operators across a contact-connectivity rebuild

matthiasneuner:
## Summary
`buildEquationSystem` was rebuilding everything from scratch every time contact connectivity changed (which facets are touching), even though a connectivity change cannot alter the mass matrix, damping, or the MPC elimination structure -- those depend on the mesh and constraints, not on which contact facets currently happen to be in contact. In a live production explicit pry-out job this rebuild runs twice per 1000 increments at ~8.2s each.

- `DofManager.refreshConstraintIndices(constraints)`: recomputes only the constraint-derived bookkeeping (constraint indices/counts/flux counts, and the VIJ block where present) using the constructor's own methods, and rebuilds the merged entity mapping as a *new* dict so anything cached on its identity (scatter template, gather plan) invalidates by construction.
- The explicit solver, on a rebuild passed `previous` (only the contact-connectivity path does that -- step start and topology changes never do), refreshes instead of reconstructing the `DofManager` and reuses M, Minv, raw M, damping rate, and the MPC transformation from `_ReusableExplicitOperators`. Gated on: matching element key set, matching MPC key set, matching nDof, and no `changematerialproperty` action in the step (mass/damping come from the materials) -- any mismatch falls back to a full rebuild. The existing `previous.U.shape` check stays.
- Bonus: the explicit solver's `DofManager` is constructed with `initializeVIJPattern=False, determiningIndexToHostObjectMapping=False` -- nothing on the explicit path reads either, and each was a full pass over ~90k elements.

## Known gap
No existing deck under `testfiles/` actually changes contact connectivity mid-run, so the solver-side reuse path is covered end-to-end only by a real pry-out run plus the new `tests/test_dofmanager_refresh.py` unit tests, not by the input-file test suite. A small sliding-contact regression deck would close that gap; none was invented for this PR.

## Verification
- pytest: 3 new tests pass; no new failures vs. a clean `next_v26.11` baseline.
- Pry-out resumed-checkpoint run, 600 increments crossing a real connectivity update, against current upstream `next_v26.11`: rebuild 3.85s -> 0.082s (47x), initial build 4.53 -> 3.59s. RF and U at the crossing increment identical to the last digit vs. the unpatched run.


matthiasneuner:
Restructured per review: force-pushed, history rewritten from 1 commit to 3.

1. Pure extraction of the mass/damping/MPC assembly into `_assembleLumpedOperators` -- zero behavior change, every original line survives verbatim.
2. The actual reuse logic on top of that extraction: removed a vacuous `nDof` guard that could never be false, documented + asserted the invariant that makes MPC-content reuse safe (every `updateConnectivity` overrider lives in `model.constraints`, not `model.multiPointConstraints`), annotated `mpcTransformation`'s real type, and added the integration test (drives the explicit solver through a real connectivity change twice -- once normally, once with reuse forced off -- and asserts U/P bit-identical).
3. The `initializeVIJPattern=False, determiningIndexToHostObjectMapping=False` change, isolated on its own: `indexToHostObjectMapping`/`getHostObjectForIndex` have zero consumers repo-wide (grepped), so this is either a first step toward removing dead code or should be flipped as a real default -- your call, didn't want to bury that decision in an unrelated commit.

Re-verified on xeon: rebuild time 8.3s -> 0.008s per rebuild (~1100x, better than the originally measured 47x), edelweiss-only/marmot/pytest all match baseline plus the new tests passing.

PR #151, opened 2026-09-19
fix(linsolve): blockamg gapCompensatedTolerance runaway + Newton-table column alignment

matthiasneuner:
## Summary

Two independent, small fixes found while diagnosing the AnchorPryOut hef70 implicit
pry-out performance investigation (see `HANDOFF_IMPLICIT_PRYOUT_LINESEARCH.md`):

- **`gapMaxFactor`** caps `BlockAMGSolver`'s true-residual gap EMA (used by
  `gapCompensatedTolerance`), which had a floor but no ceiling. A single solve that
  fails to converge reports a measured gap many orders of magnitude above the
  documented normal range (1.0-3.5); the EMA carries that forward, and every
  subsequent solve's first-pass tolerance collapses toward its own `1e-14` floor
  with no self-correction -- a positive-feedback runaway. Observed live: 5+
  consecutive solves each burning ~2400 outer iterations while the achieved true
  residual *worsened* monotonically (7.7e-3 -> 5.1e-2 -> 3.2e-1 -> 7.5e-1 -> 9.4e-1).
  Default `gapMaxFactor=1e3` is comfortably above the documented normal range and
  is a no-op unless a solve actually goes pathological.

- **Newton-table column alignment**: the per-field header labels (`||R||∞`,
  `||ddU||∞`) were left-justified in an ad hoc 10-wide slot that never matched the
  actual 12-wide value+marker cells the data row prints. The "linear solve" column
  added by an earlier refactor (5d9d5e36, merging blockamg's per-solve diagnostics
  into the Newton table) inherited the same mismatched format, plus no visual
  separation from the real fields' columns. Fixed by centering each label over its
  own 12-wide cell (matching the data format exactly) and adding a 1-space gap
  before the linear-solve group -- capped at 1 space because the Journal wraps
  level-2 messages at 76 characters and this model's row is already at 75 without
  any gap at all; a wider gap was tried first and wrapped the row's tail onto its
  own line, which looks considerably worse than tight spacing.

Both were found and fixed live against a running LEO5 job via a bind-mounted
pure-Python patch (no container rebuild needed), then properly rebased and
committed here against current `next_v26.11`.

## Test plan

- [x] `py_compile` clean on all touched files
- [x] Diffed the deployed bind-mounted files directly against the container's own
      baked-in originals to confirm only the intended lines differ
- [x] Verified live on LEO5: `gapMaxFactor` present and effective (confirmed via
      `inspect.signature`); Newton-table rows render at exactly 100 chars (no wrap)
      across multiple increments on both a plain and an AMR-enabled run, both
      correctly resumed from restart checkpoints through this change
- [ ] No new automated test added for the EMA cap specifically (the failure mode
      needs a solve that goes pathological, which is awkward to construct as a
      unit test) -- flagging for reviewer judgment on whether one is expected


edelweissfe/linsolve/blockamg/blockamg.py
```diff
@@ -340,6 +340,18 @@ class BlockAMGSolver(LinearSolver):
         per unit of *simulated time* rather than per increment, the per-increment saving is largely
         or wholly given back. ``gapSafetyFactor`` is the knob between the two regimes (smaller =
         more accurate = closer to the default's Newton behaviour); the sweet spot is unexplored.
+
+        **Bug found and fixed 2026-09-19 (`gapMaxFactor`):** the EMA in :meth:`__call__` that learns
```
matthiasneuner: WTF IS THIS SHIT? THIS DOES NOT BELONG TO A DOCSTRING

matthiasneuner: Fair, cut it -- turned it into a normal 3-line parameter doc entry, and trimmed the same over-explaining pattern from the other inline comments across all three files (c54de42d). Bug narrative belongs in the commit message, not the docstring.

